PR #1, opened 2020-02-10
responsive-grid-project

minwwgo:


index.html
```diff
@@ -1,20 +1,154 @@
...
 
-</body>
+    <header>
+      <nav class="container nav-bar">
+        <div class="row head-logo">
+          <h3 class="logo">sumo</h3>
+        </div>
+
+        <ul class="row nav-list">
+          <li class="sm-col-4 md-col-4 lg-col-4">
```
AlfioEmanueleFresta: Always start from `col-N`, which is equivalent to `xs-col-N`. Otherwise you will be missing any styling for small screens (e.g. smartphones). 

Additionally, you only need to include larger-screen classes if they change. Here's a few examples:
```html
<li class="col-4">            <!-- col-4 for all screen sizes -->
<li class="xs-col-4">         <!-- same as the one above -->
<li class="col-12 md-col-6">  <!-- full-width on small screens, half-width on medium-size and above -->
```

index.html
```diff
@@ -1,20 +1,154 @@
...
+      <section class="container about">
+        <div class="row hero-content-row">
+          <div class="sm-col-12 md-col-12 lg-col-12">
+            <img
+              class="hero-content-img"
+              src="/images/fruit-vertical-three.jpeg"
+            />
+          </div>
+          <div class="sm-col-12 md-col-12 lg-col-12">
+            <h2 class="sm-col-12 md-col-12 lg-col-12">about</h2>
```
AlfioEmanueleFresta: Use a `row` around any `col-*` elements.

index.html
```diff
@@ -1,20 +1,154 @@
...
+            <label for="email">Enter Your Email *</label><br />
+            <input type="text" id="email" placeholder="   Email" /><br />
+            <label for="phone">Enter Your Phone *</label><br />
+            <input type="text" id="phone" placeholder="   Phone" /><br />
+            <label for="message">Enter Your Message here *</label><br />
+            <input
+              class="message-box"
+              type="text"
+              id="message"
+              placeholder="   Message"
```
AlfioEmanueleFresta: If you want a multi-line input box, you can look into `<textarea>` instead of `<input>`.

Also, if you want to style a form element, use CSS - prefer `text-align` to spaces! There's even [a selector](https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder) which allows you to change the style of the placeholder text.

index.html
```diff
@@ -1,20 +1,154 @@
...
+            </div>
+              
+            </div>
+            
+          </form>
+        </div>
+      </section>
+    </main>
+    <footer>
+      <section class="container">
```
AlfioEmanueleFresta: You don't need `<section>` in the `<footer>`. You can use a plain `<div>` instead.

AlfioEmanueleFresta:
Well done @minwwgo. Your code is also easy to read - great indentation! I've left a few comments above.

PR #3, opened 2020-02-12
Update index.html

TamBahri:


brightpixels:
Very good start Tam. Well done for start with mobile

index.html
```diff
@@ -6,15 +6,185 @@
...
+
 
-    <!-- Add your markup here -->
+            </div>
+     </div> 
 
+    </div>
+    <div>
+        <div class="col-12 md-col-5">
+            <div class="form">
```
brightpixels: To make your form more accurate, wrap it with a `form` tag. So all the fields and submit button should be part of a single `form` tag. See example here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form

index.html
```diff
@@ -6,15 +6,214 @@
...
+                
+            </div>
+
+        </section>
+       
+      
+          
+    </header>
+
+    <section class="container">
```
LucyMac: Consider moving everything down one level so that you:
- first set your background image (without a `container` around it so it goes all the way to the edges of the screen). This means moving your `hero` class to the top; 
- then, you add the `container` _inside_ of that (as a child of the element that has the background image)
- then, `row`
- then `col-12`

So basically the same as you've done except you would be moving the `hero` class to wrap everything else. This will have a much better visual effect 😄

index.html
```diff
@@ -6,15 +6,214 @@
...
+                </div>
+                
+
+
+            </div>
+     </div> 
+
+    </div>
+    <div>
+        <div>
```
LucyMac: Unnecessary `div`s. Remove and start with `<div class="row">` straight away.

PR #4, opened 2020-02-12
Abdullah Celen / Plant shop website

LucyMac:


index.html
```diff
@@ -7,14 +7,181 @@
...
+                        <select id="country" name="country">
+                          <option value="australia">Australia</option>
+                          <option value="canada">Canada</option>
+                          <option value="usa">USA</option>
+                        </select>
+                    
+                        <label for="subject">Subject</label>
+                        <textarea id="subject" name="subject" placeholder="Write something.." style="height:200px"></textarea>
+                    
+                        <input type="submit" value="Submit">
```
amcraethanan: Instead of input this could be a button html element with bootstrap classes

index.html
```diff
@@ -7,14 +7,181 @@
...
+        
+   <!-- MAP -->        
+            <div class="row-3">
+                <div class="col-12 md-col-6">
+                    <div class="card">
+                        <img src="/images/map.PNG" alt="hand">
+                    </div>   
+                </div> 
+                <div class="col-12 md-col-6"> 
+                <div class="row">
```
amcraethanan: Careful with your identation here - the row class is a child of your col class here

index.html
```diff
@@ -7,14 +7,181 @@
...
+          
+                        <div class="col-6">
+                            <h4>Address: </h4><h5> <br> Lorem ipsum dolor sit amet consectetur adipisicing elit. Omnis sunt placeat at odit, architecto corrupti nemo molestiae! Modi vero eius amet. Laborum assumenda, ullam beatae facere ea enim harum voluptatum.</h5>
```
amcraethanan: Instead of h5 , this should be a p element instead with a class applied

index.html
```diff
@@ -7,14 +7,181 @@
...
+                        <div class="col-6">
+                            <h4>jdf lorem10 dsfd df sdf dfdsf dfs sdf s s sdf sdfs fsd fs </h45>
+                        </div>
+                    </div>
+                </div>
+            </div>
+    </div>
+</section>
 
+<section>
```
amcraethanan: Instead of making a new section, this should be a `footer` instead

amcraethanan: you can then chain the footer element with mutiple classes, so the "footer" and "container" class can be applied to one html element

index.html
```diff
@@ -7,14 +7,181 @@
...
+                <div class="col-12 md-col-4">
+                    <div class="content">
+                        <h3>By:</h3> 
+                        <span class="copy">&copy; Abdullah Celen</span>
+                    </div>
+                </div>
+                <div class="col-12 md-col-4">
+                    <div class="content">
+                        <h3>Join us on</h3>
+                        <p>plant <a href="#"></a> <br>garden <a href="#"></a></p>
```
amcraethanan: You've got a couple of empty `a` elements here - if you're not going to use them, please remove them so that only used html elements are part of your site

css/main.css
```diff
@@ -1,3 +1,201 @@
...
+    margin-right: 0px;
+    margin-left: 0px;
+  }
 
-/* Dont' forget to link this file to your HTML in the <head> */
+  .about-col-6 {
+    width: 100%;
+    position: relative;
+    min-height: 1px;
+    padding-right: 0px;
```
amcraethanan: if not definied, the padding (like margin) will be set to 0 - so this shouldn't be needed here

css/main.css
```diff
@@ -1,3 +1,201 @@
 /* Add your custom CSS here */
+@import url('https://fonts.googleapis.com/css?family=PT+Sans&display=swap');
+body {
+    font-family: 'PT Sans', sans-serif;
+    background: rgb(219, 218, 218);
+    color: sienna;
+}
+/* custom gridfor about-us */
+.row-2 {
```
amcraethanan: the bootstrap grid classes should already provide this as "row" - see the comment made on your index.html

css/main.css
```diff
@@ -1,3 +1,201 @@
...
+}
+
+h2{
+    
+    text-align: center;
+    color: rgb(129, 48, 11);
+    padding-top: 10px;
+}
+
+h3{
```
amcraethanan: Remember to format your css as you go with the right spacing eg `h2 {`

css/main.css
```diff
@@ -1,3 +1,201 @@
...
+}
+
+h4{
+    text-align: center;
+    padding: 10px;
+}
+h5{
+    padding: 5px 10px ;
+}
+P{
```
amcraethanan: this should be lowercase `p` when defining styles for the `p` element

css/main.css
```diff
@@ -1,3 +1,201 @@
...
+    display: inline-block;
+    margin-left: 1.563rem;
+}
+a:hover{
+    color: black;
+}
+
+.hero{
+    
+    height: 50vh;
```
amcraethanan: By defining the height, when the screen gets to very small, the button on the html page overflows over the div with the hero class - do we need to define the height here as `50vh`?

PR #5, opened 2020-02-13
Sc responsive grid project

SCavus:
It would be nicer if I could add a hamburger menu on the mobile version.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -16,18 +16,6 @@ body {
     align-items: flex-start;
 }
 
-@media (min-width: 768px) {
-    .navlist {
-        display: flex;
-        flex-direction: row;
-        margin-right: -15px;
-        margin-left: -15px;
-    }
-    li {
-        text-align: center;
-    }
-}
-
 .job {
     text-align: center;
     margin-top: 0;
@@ -69,36 +57,15 @@ li {
     margin-left: -15px;
 }
 
-@media (min-width: 768px) {
-    .intro {
-        min-height:500px ;
-    }
-    .ambition {
-        font-size: small;
-        margin: 4rem 0 0 0;
-    }
-}
-
-.ambition {
+#ambition {
     font-size: small;
     margin: 4rem 0 0 0;
 }
 
-.succees {
+.success {
     font-size: 3rem;
     font-weight: bolder;
     margin: 0.5rem 0 0 0;
-}
-
-@media (min-width: 768px) {
-   .succeed {
-    font-size: 7rem;
-    font-weight: bolder;
-    margin: 0.5rem 0 0 0;
-    }
-    button {
-        margin-top:2rem;
-    }
 }
 
 button {
@@ -110,6 +77,8 @@ button {
 .about {
     margin: 0;
     background-color: rgb(235, 231, 217);
+    display: flex;
+    flex-direction: column;
 }
 
 .photo {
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -5,14 +5,135 @@
...
+            </ul> 
+        </nav>
+        <div class="row col-3">
+            <i class="col-4 fa fa-user-circle-o"></i>
+            <p class="col-8 login">Log in</p>
+        </div>
+    </header>
+    <section class="intro">
+        <div>
+            <p class="ambition">AMBITION IS THE FIRST STEP TOWARDS</p>
```
ChrisOwen101: I think it makes more sense to use an `id` here because there is only one of the type of text

index.html
```diff
@@ -5,14 +5,135 @@
...
+                </div>
+            </div>
+        </div>
+        <div class="photo col-12 md-col-6"></div>
+    </section>
+    <section class="row pricing">
+        <div class="md-col-2"></div>
+        <div class="md-col-8">
+            <div class="heading">
+                <p>HOW I CAN HELP YOU</p>
```
ChrisOwen101: I think it makes more sense to make these `h1`, `h2` or `h3` tags and then style them all together. 

Using the `H` tags makes the HTML easier to read at a glance because you can tell what is meant to be a title and what is meant to be normal text. This is a really small problem though!

css/main.css
```diff
@@ -1,3 +1,264 @@
...
+footer {
+    background-color: skyblue;
+}
+
+.contact {
+    font-size: 0.8rem;
+    margin: 1rem 0 0 1rem;
+}
+
+@media (min-width: 768px) {
```
ChrisOwen101: Generally, it's better to group all of the media query css together so that you can tell more easily what changes as the screen resizes. For example,

```
@media (min-width: 768px) {
    .form {
        display: flex;
        flex-direction: column;
        padding-top: 3rem;
        padding-left: 3rem;
     }

    .contact {
        margin: 3rem 0 0 3rem;
    }

    h2 {
        margin-top: 3rem;
    }

    h4 {
        padding: 0 5rem 2rem 5rem;
    }
}
```
Could all go inside the same media query

css/main.css
```diff
@@ -1,3 +1,264 @@
...
+}
+
+.succees {
+    font-size: 3rem;
+    font-weight: bolder;
+    margin: 0.5rem 0 0 0;
+}
+
+@media (min-width: 768px) {
+   .succeed {
```
ChrisOwen101: I think this should be `succees` so that it works with your HTML. At the moment, this does nothing

css/main.css
```diff
@@ -1,3 +1,264 @@
...
+    align-items: center;
+    margin-right: -15px;
+    margin-left: -15px;
+}
+
+@media (min-width: 768px) {
+    .intro {
+        min-height:500px ;
+    }
+    .ambition {
```
ChrisOwen101: This can be removed as it's the same as the `.ambition` outside of the media query

index.html
SCavus: I have replaced it with id.

css/main.css
SCavus: I put all media query CSSs in one block.

css/main.css
SCavus: Yes, actually it should be "success".  I have corrected all of them.

css/main.css
SCavus: You are totally right. I was trying to change the font size for mobile and the desktop. After I copied the whole .ambition block, forgat to correct it. So I didn't remove it, instead, corrected it. It has now different font sizes on mobile and desktop.

PR #6, opened 2020-02-13
Responsive grid project nader akhgari

naderakhgari:


index.html
```diff
@@ -6,15 +6,59 @@
...
+            <h2 class="hide"> Call Us Now 123-456-7890</h2>
+            <div class="right-side sm-col-6">
+                <img src="../img/user-logo.png" class="sm-col-1">
+                <a href="https/www.google.com" class="login-font">LogIn</a>
+            </div>
+        </div>
+    </nav>
+    <section class="first-section sm-col-12">
+        <!-- <img src="../img/first-sec-img.jpg" class="first-section-img col-12"> -->
+        <div class="SM-COL-4">
```
amcraethanan: be careful of your formatting here, this class should be "sm-col-4"

index.html
```diff
@@ -6,15 +6,59 @@
...
+        </div>
+        <div class="sm-col-6 no-padding right-side">
+            <h2 class="hide"> Call Us Now 123-456-7890</h2>
+            <div class="right-side sm-col-6">
+                <img src="../img/user-logo.png" class="sm-col-1">
+                <a href="https/www.google.com" class="login-font">LogIn</a>
+            </div>
+        </div>
+    </nav>
+    <section class="first-section sm-col-12">
```
amcraethanan: This is now the start of your main part of your website - wrap this in the semantic `<main>` html element to indicate that this is the case.

index.html
```diff
@@ -6,15 +6,59 @@
...
+    
+
     <!-- Add a link to your custom CSS here -->
 
     <title>Responsive grid project</title>
-</head>
+    </head>
 
-<body>
+<body class="sm-col-12 ">
```
amcraethanan: Remember when implementing the 12 grid system to first have a html element with class set to "container", a child element then set to "row" (so that flexbox is implemented) and then use "col-" classes for your child elements.

css/main.css
```diff
@@ -1,3 +1,275 @@
...
+}
+.left-side{
+    display: flex;
+    justify-content: flex-start;
+}
+.first-section{
+    background-color: skyblue;
+    background: url("../img/first-sec-img.jpg") no-repeat center;
+    color: snow;
+    height: 100vw;
```
jamesgorrie: Why do you want this to be 100v__w__ height?

index.html
```diff
@@ -6,15 +6,130 @@
...
-</head>
+    </head>
 
-<body>
+<body class="col-12 sm-col-12 ">
 
     <!-- Add your markup here -->
+    <nav class="nav-content col-12 sm-col-12 lg-col-12 no-padding" >
+        <div class=" no-padding col-6 sm-col-6 lg-col-6 left">
+            <div class="left-side col-12 sm-col-12 lg-col-12 no-padding">
```
jamesgorrie: It looks like we have a `col-12` in a `col-6` in a `col-12`...

The grid system is meant to be `.container => .row => .col-number` - it might be worth looking at simplifying this.

PR #9, opened 2020-02-13
Curent Version-grid/tayfuntayfun

tayfuntayfun:
i am keeping to work on my page but this is the current version. Feedback would be much appreciated.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -3,8 +3,8 @@
 /* Dont' forget to link this file to your HTML in the <head> */
 
 .container {
-    border: 2px solid rgb(58, 2, 58);
-    background-color: #ebe8f0;
+    /* border: 2px solid rgb(58, 2, 58);
+    background-color: #ebe8f0; */
 }
 
 header {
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -5,16 +5,131 @@
...
+    <link rel="stylesheet" type="text/css" href="css/main.css">
     <!-- Add a link to your custom CSS here -->
-
+    
     <title>Responsive grid project</title>
 </head>
 
 <body>
+    <div class="container">
+        <header class="row"><!--Start Row 1-->
```
ChrisOwen101: I think these comments make the code a little hard to read but if they help you then 👍

Maybe remove them when you're done with the homework

css/main.css
```diff
@@ -1,3 +1,87 @@
...
+    margin: 0;
+} 
+
+.navbar{
+justify-content: space-around;
+background-color: blueviolet;
+}
+
+.list-elements{
+list-style-type: none;
```
ChrisOwen101: These can be formatted automatically by turning on this setting:
https://stackoverflow.com/questions/39494277/how-do-you-format-code-on-save-in-vs-code

ChrisOwen101:
Great work on this so far Tayfun! If you have any more question then message me on Slack or on the #london-class6 channel :)

Keep it up!

index.html
tayfuntayfun: Thanks Chris, you are right there are a lot. I have removed unnecessary comments and now working on media query. I am actually thinking something but not sure how to do. my header includes a logo a navbar and contact/sign in icons. I want navbar come under the header and be in column direction. BAsically, once the screen size decreases to a phone level, the navbar is to be removed from the header (navbar's container) and becomes another container after the header. By doing that, I can (on a phone screen) show header on top, navbar elements in column and the hero image and so on...Researching about this I found that I can disable some parts of the html but this is a bit tricky. I am leaving this for later unless you have a basic solution

css/main.css
tayfuntayfun: Hi Chris, forgot to populate here that I did automatic format for the script.

PR #11, opened 2020-02-14
My first pull

AlexandruBudaca:


index.html
```diff
@@ -37,7 +37,7 @@ <h2>Rele<span>V</span>ance</h2>
             </ul>
```
sophiedermaux: Where is the opening tag of this <ul>? if it isn't used it should be deleted

index.html
```diff
@@ -93,16 +93,51 @@ <h3>Aknknsadk</h3>
         </div>
     </div>
     <br>
-    <div class="container">
-        <div class="row ">
+    <!--Second row of products-->
+    <div class="container products-2nd">
+        <div class="row products">
+            <div class=" box col-6 col-4">
```
sophiedermaux: remove the space at the beginning of this class before the word box

index.html
```diff
@@ -93,16 +93,51 @@ <h3>Aknknsadk</h3>
...
+                <div>
+                    <img src="/img/shoes.jpeg" alt="">
+                    <h3>Aknknsadk</h3>
+                    <p>Price: $100</p>
+                </div>
+            </div>
+        
+            <div class=" box col-6 col-4">
```
sophiedermaux: same feedback as above

index.html
```diff
@@ -115,24 +150,128 @@ <h3>About us</h3>
...
                 <h1>BUY ONLINE <br> 
                     NOW & GET<br>
                      10% OFF!</h1>
             </div>
+        </div>
+    </div>
+<!--form-->
+    <div class="container section-form">
+        <h2>Contact Us</h2>
+        <div class="row ">
```
sophiedermaux: remove the space at the end of the tag row

index.html
```diff
@@ -115,24 +150,128 @@ <h3>About us</h3>
...
+
+    </div>
+
+<!--Footer-->
+    <div class="container section-footer">
+        <div class="row">
+
+            <div class=" box col-12 xl-col-3 ">
+                <div >
+                    <h3>Our Store</h3>
```
sophiedermaux: You reference h3 but don't include it in the css? if it is used often could be worthwhile to add your own CSS for it

sophiedermaux:
Great job so far Alexandru! I recommend you check for a couple of things 1) ensure there is no space after the words in your tags, you can see in tags such as row and box 2) you use one big container and then have sub containers, check if the big container is necessary.

css/main.css
```diff
@@ -1,3 +1,391 @@
...
+    color: red;
+}
+
+.grid-img {
+    position: relative;
+}
+.sale {
+    background-color: rgba(0,0,0, .3);
+    width: 100%;
+    position: absolute;
```
LucyMac: Be careful with the use of position: absolute - it is causing your elements to overlap each other. 
Add images first and once the elements on the page have their own height, you'll be able to display text on top of them if needed.

css/main.css
```diff
@@ -1,3 +1,391 @@
...
+    flex-direction: column;
+    width: 100%;
+    justify-content: center;
+    align-items: center;
+    
+}
+.header .menu-icon {
+    cursor: pointer;
+    display: block;
+    float: right;
```
LucyMac: Don't use float rules, use flexbox instead.

index.html
```diff
@@ -6,13 +6,273 @@
...
+                <li><a href="">Home</a></li>
+                <li><a href="">Shop</a></li>
+                <li><a href="">About</a></li>
+                <li><a href="">Blog</a></li>
+                <li><a href="">Contact</a></li>
+            </ul>
+    </header>
 
+<!--main-header-->
+    <div class="container">
```
LucyMac: could you use the `main` tag here?

index.html
```diff
@@ -6,13 +6,273 @@
...
+                </div>
+            </div>
+                
+              </form>
+        </div>
+
+    </div>
+
+<!--Footer-->
+    <div class="container section-footer">
```
LucyMac: Use the `footer` tag!

css/main.css
```diff
@@ -1,3 +1,391 @@
...
+   }
+   .section-about .col-3 {
+    width: 25%;
+    }
+    .products .col-4 {
+        width: 33.333333%;
+    }
+    .discount .col-6 {
+        width: 50%;
+    }
```
LucyMac: Never set your own widths for `col-...` classes. That's the whole reason they exist - they have already been written!

PR #13, opened 2020-02-15
Responsive-Grid/Portfolio/Mo

mnashawati:


sophiedermaux:
Great job Mo!! You can see you experimented with the columns and that is great. I would recommend in just ensuring your indentation is clear and always in line for reviewing

css/main.css
```diff
@@ -1,3 +1,200 @@
...
+    display: flex;
+    justify-content: center;
+}
+
+header ul li {
+    list-style-type: none;
+    padding: 1rem;
+}
+
+header ul li a {
```
theseafaringturtle: It's probably best to use a class for this for maintainability

css/main.css
```diff
@@ -1,3 +1,200 @@
...
+header ul li a:hover {
+    opacity: 1;
+}
+
+h2 {
+    padding-top: 1rem;
+}
+
+.hero {
+    background-image: url('../imgs/photo-1490117874548-e35a2286fd89.jpeg');
```
theseafaringturtle: ../ is not necessary here since / starts from the site's root folder

index.html
```diff
@@ -6,14 +6,100 @@
...
+
+        <section class="third-bg-img">
+        </section>
+
+        <section id="contact" class="contact">
+            <h1>Contact.</h1>
+            <div class="container">
+                <form class="row">
+                    <label class="col-10 sm-col-10 md-col-6 lg-col-5 info-label" for="name">Enter Your Name</label>
+                    <input class="col-10 sm-col-10 md-col-6 lg-col-5" type="text" id="name" name="name" value="Name" required><br>
```
theseafaringturtle: I recommend using placeholder= rather than value= in inputs

mnashawati:
> Great job Mo!! You can see you experimented with the columns and that is great. I would recommend in just ensuring your indentation is clear and always in line for reviewing

Thank you for the feedback Sophie! Could you please explain what exactly I can do about the indentation? I use "Format Document with Beautify" for HTML in VS Code so it is automatic.

index.html
mnashawati: Thank you for the note Alessandro :) Noted and done!

PR #16, opened 2020-02-15
done

sabia-m:
Best I could do, but my understanding of CSS Grid is very weak, so I'm aware it needs lots of improvement.


css/main.css
```diff
@@ -1,3 +1,99 @@
...
+    font-size: 1.5vw;
+    background-color: black;
+    color: white;
+}
+
+.row3 .col-6 {
+    height: 400px;
+}
+
+.row3 .left {
```
jamesgorrie: In stead of overriding these grid classes, you could create some new classes that are more specific to this style. e.g. `.light-blue`.

This will make it easier for people in the future to know how this works.

css/main.css
```diff
@@ -1,3 +1,99 @@
...
+}
+
+.book-now {
+    text-align: center;
+    font-size: 1.5vw;
+    background-color: black;
+    color: white;
+}
+
+.row3 .col-6 {
```
jamesgorrie: See this comment
https://github.com/CodeYourFuture/responsive-grid-project/pull/16/files#r379918900

css/main.css
```diff
@@ -1,3 +1,99 @@
...
+    background-color: rgb(191, 211, 248);
+}
+
+.read-more {
+    background-color: teal;
+    color: white;
+    font-size: 1.5vw;
+}
+
+.row3 .right {
```
jamesgorrie: See this comment
https://github.com/CodeYourFuture/responsive-grid-project/pull/16/files#r379918900

index.html
```diff
@@ -5,16 +5,83 @@
...
 </head>
-
 <body>
+<div class="container">
+    <div class="row"> <!--header-->
+        <header>
+            <div class="col-6">
+            <h1 class="logo-title">ALLAN JOHNSON</h1>
+            <h3 class="logo-sub">Personal Life Coach</h3>
+        </div>
```
jamesgorrie: Is this the right closing tag?

index.html
```diff
@@ -5,16 +5,83 @@
...
 
     <title>Responsive grid project</title>
 </head>
-
 <body>
+<div class="container">
+    <div class="row"> <!--header-->
+        <header>
+            <div class="col-6">
+            <h1 class="logo-title">ALLAN JOHNSON</h1>
```
jamesgorrie: Tiny point, having all caps can cause issues with accessibility:
https://github.com/humanmade/hm-pattern-library/issues/75#issuecomment-235672885

Not a biggy, but you can have them in normal case here, `Allan Johnson`, and create a `.caps` class in the CSS with `text-transform: uppercase`.

https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform

index.html
```diff
@@ -5,16 +5,83 @@
...
+            <h1 class="box">ONE ON ONE COACHING</h1>
+            <p class="coaching-info">1 hr<br>£40</p>
+            <button class="book-it" type="button">Book It</button>
+        </div>
+    </div>
+    </div>
 
+    <div class="row"> <!--Footer & Form-->
+        <footer>
+            <div class="row-5">
```
jamesgorrie: What does this class do?

index.html
```diff
@@ -5,16 +5,83 @@
...
-    <link rel="stylesheet" type="text/css" href="css/grid.css">
+    <link rel="stylesheet" type="text/css" href="css/grid.css"> 
+    <link rel="stylesheet" type="text/css" href="css/main.css">
     <!-- Add a link to your custom CSS here -->
 
     <title>Responsive grid project</title>
 </head>
-
 <body>
+<div class="container">
```
LucyMac: This is a good use of the container as it 'contains' (literally) your entire page.
However, you will notice that because of this, your page elements don't span to the edges of the screen. In a lot of cases, depending on the design, you'll want the background colour of your nav and background image of your hero (for example), to grow as wide as the screen, but for _the content_ (i.e. the text) to be contained.
To achieve this, instead of wrapping the whole thing in a container, you can use the container inside individual sections. For example for a hero you'd do something like this:
```
    <section class="my-full-width-background-class">
        <div class="container">
            <div class="row">
                <div class="col-12">
                    <h4>This is the hero!</h4>
                </div>
            </div>
        </div>
    </section>
```
Note how we set the background image in the outer parent, and then _inside_ of that we put the container, to container the text etc.

index.html
```diff
@@ -5,16 +5,83 @@
...
+    <div class="row"> <!--header-->
+        <header>
```
LucyMac: Avoid having an element (like your `<header>` here) between the `row` and the `col-...`. The reason for this is that the `row` class has `display: flex` on it, and flexbox only works on immediate, direct children elements.
So here you could get rid of the `div` on line 16, and move your `row` class to be on the `<header>`.

index.html
```diff
@@ -5,16 +5,83 @@
...
+            <div class="col-6">
+            <h1 class="logo-title">ALLAN JOHNSON</h1>
+            <h3 class="logo-sub">Personal Life Coach</h3>
+        </div>
+            <nav>
+                <div class="col-6">
```
LucyMac: Similar comment as above - move the `col-...` class to be directly on the `nav` so that it is a direct child of the `row`. And get rid of the div on line 23 as you won't need it anymore.

index.html
```diff
@@ -5,16 +5,83 @@
...
+                    <li>Contact</li>
+                    <li>Log In</li>
+               </ul>
+        </nav>
+        </header>
+    </div>
 
-    <!-- Add your markup here -->
+    <div class="row"> <!--hero-->
+        <div class="col-12 hero">
```
LucyMac: Good use of the grid here

index.html
```diff
@@ -5,16 +5,83 @@
...
+            <p class="ambition">Ambition is the first step towards</p>
+            <p class="success">SUCCESS</p>
+            <button class="book-now" type="button">Book Now</button>
+        </div>
+    </div>
+    <div class="row"> <!--About Me-->
+       <div class="row3">
```
LucyMac: Less good here. 2 things:
1) As mentioned above, don't add any elements between the `row` and the `col-6`. `col-6` should be a direct child of `row`.
2) Avoid re-using class names that are used in the grid system to avoid confusion. If they're your own classes, give them their own name 🙂

LucyMac: If you do this (and by doing that, get rid of `row3`, it will fix the issue you have with both left and right sections stacking vertically.

index.html
```diff
@@ -5,16 +5,83 @@
...
+    <div class="row"> <!--How I Can Help-->
+        <div class="col-12 help">
+            <h1>How I Can <br>Help You</h1>
+        <div class="coaching">
+            <h1 class="box">ONE ON ONE COACHING</h1>
+            <p class="coaching-info">1 hr<br>£40</p>
+            <button class="book-it" type="button">Book It</button>
+        </div>
+    </div>
+    </div>
```
LucyMac: Watch your indentation, it makes it hard to follow.

LucyMac: This 'How can I help' section is also a good candidate for what I mentioned above: once you've removed the overall `container`, you can add a `container` inside this section, after the background images has been set (which should be full-width), and before the `row` and the `col-12`.
(see code snippet shared above)

index.html
```diff
@@ -5,16 +5,83 @@
...
+                <div class="col-6 right">
+                    <form class="footer-form">
+                        <label for="name">Name</label>
+                        <input type="text" id="name" name="name">
+                        <label for="email">Email</label>
+                        <input type="text" id="email" name="email">
+                        <label for="message">Message</label>
+                        <input type="text" id="message" name="message">
+                        <input type="submit" value="Submit">
+                    </form>
```
LucyMac: Can you add some styling to your inputs in your CSS, so it matches the design? 🙂

PR #17, opened 2020-02-15
Responsive grid abbas mokhtari

abbasmokhtari:
I couldn't fix the issue with submit button and social media icons


css/main.css
```diff
@@ -1,3 +1,83 @@
...
+ul {
+    list-style: none;
+}
+
+.nav-bar a {
+	text-decoration: none;
+	color: rgb(54, 54, 54);
+}
+
+img {
```
jamesgorrie: This will affect all images that are put in the HTML, is that what we want?

css/main.css
```diff
@@ -1,3 +1,83 @@
...
+    width: 100vw;
+}
+
+
+.section-a h2, .section-b h2, .section-c h2 {
```
jamesgorrie: You could create a new class called something like: `.centered-heading` and just apply that to the `h2` elements in the HTML.

index.html
```diff
@@ -7,14 +7,141 @@
...
     <title>Responsive grid project</title>
 </head>
 
 <body>
-
     <!-- Add your markup here -->
 
+    <!-- Header and nav bar -->
+    <header id="header">
+        <h1 class="md-col-3">
```
jamesgorrie: I think the structure from the [CSS](https://github.com/CodeYourFuture/responsive-grid-project/blob/master/css/grid.css) should be:

```html
<div class="container">
<div class="row">
<!-- put columns here -->
</div>
</div>
```

index.html
```diff
@@ -7,14 +7,141 @@
...
+                        allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"
+                        allowfullscreen></iframe>
+
+                    <iframe class="md-col-4"  src="https://www.youtube.com/embed/nfVUk5xU1gM" frameborder="0"
+                        allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture"
+                        allowfullscreen></iframe>
+                </div>
+            </div>
+
+            <img src="/img/jackson-hayes-UU77aJC-Wb0-unsplash.jpg" alt="mobile edit">
```
jamesgorrie: Does the al text here describe the image?

PR #20, opened 2020-02-16
RC rough version

279Roxana:


index.html
```diff
@@ -5,15 +5,198 @@
...
+
+                <div class="col-2 login">
+                    <i class="fa fa-user-circle"></i>
+                    <a href="#">Log In</a>
+                </div>
+
+            </nav>
+
+        </div>
+        </div>
```
amandaeareias: Be careful with extra closing tags that have not been opened (or vice-versa). Important to keep indentation always correct to make it easier to spot them

css/main.css
```diff
@@ -1,3 +1,187 @@
...
+        display: flex;
+        flex-direction: column;
+        align-content: center;
+        max-width: 100%;
+        top: 2.5rem;
+    }
+    
+    
+    .section2-image {
+        width: 100%;
```
amandaeareias: use the widths specified in grid.css

PR #21, opened 2020-02-16
Ellie-first pull

ellietms:
I had few problems with my ham menu,I was just confused:)


css/main.css
```diff
@@ -1,3 +1,254 @@
...
+}
+.col-3 {
+  width: 25%;
+}
+.col-2 {
+  width: 16.666666%;
+}
+.col-1 {
+  width: 8.333333%;
+}
```
LucyMac: You don't need to write this - that's what grid.css is for! It provides all the grid classes pre-written for you. You just need to import it in the `<head>` of your index.html and they will be ready to use!

css/main.css
```diff
@@ -1,3 +1,254 @@
...
+
+
+
+.container {
+  max-width: 100%;
+  padding-right: 15px;
+  padding-left: 15px;
+  margin-right: auto;
+  margin-left: auto;
+}
```
LucyMac: Don't re-write these!

index 21.48.04.html
```diff
@@ -0,0 +1,69 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <link rel="stylesheet" type="text/css" href="main.css">
```
LucyMac: Import grid.css here

index 21.48.04.html
```diff
@@ -0,0 +1,69 @@
...
+        <li>co-founded Apple Computers</li>
+      </ul> 
+    </div> 
+    <div class="row">
+    <div class="md-col-2"> 
+      
+        Home
+      
+      <div class="md-col-2">
+        <div class="row">
```
LucyMac: `row` always comes before the `col-...` class

css/main.css
```diff
@@ -1,3 +1,254 @@
...
+
+.logo_img{
+  background: url("steve-jobs-nav.jpg");
+
+}
+
+.hero_image{
+  display: flex;
+  background: url("steve_jobs.jpg");
```
LucyMac: try `background: url(../steve_jobs.jpg);`

css/main.css
```diff
@@ -1,3 +1,203 @@
...
+  background-color: #466381;
+  cursor: pointer;
+  margin-top: 5px;
+}
+.about_me_img{
+  display: flex;
+  justify-content: flex-start;
+  text-align:flex-start;
+  background-repeat: no-repeat;
+  background-size: cover;
```
LucyMac: These rules won't work because your image is not a background-image. 
The image for this is actually inside an `img` tag in your HTML, and it's getting distorted at the moment. Remove that and set the image to be a background-image, like you did for the Steve Jobs picture. That way, it will cover the whole space and scale much better.

css/main.css
```diff
@@ -1,3 +1,203 @@
-/* Add your custom CSS here */
```
LucyMac: The browser adds a default 8px margin on the body which is why there is a white edge around your page. 
Add this to remove it:
```
body {
   margin: 0;
}
```

index 21.48.04.html
```diff
@@ -0,0 +1,69 @@
...
+    </title>
+  </head> 
+  <section class="container">
+  <header class="header_customize">
+    <div class="part_one_nav">
+      <ul>
+        <li>Steve Jobs</li>
+        <br>
+        <li>co-founded Apple Computers</li>
+      </ul>
```
LucyMac: Is this really a list? Just use a `section` with some heading tags.

index.html
```diff
@@ -5,16 +5,158 @@
...
+    <main>
+        <section class="hero_image section_one">
+            <h1 class="font_anton">
+                SUCCESS
+            </h1>
+            <button class="button col-2">
+                Book Now
+            </button>
+        </section>
+        <section>
```
LucyMac: If you split the text part below and the image part (with the product image) into 2 different sections, you could add a `container` on your text section to 'contain' the text to the center of the screen on large screens. At the moment, the text goes all the way to the edges which looks a bit much on desktop.

index.html
```diff
@@ -5,16 +5,158 @@
...
 
-    <!-- Add your markup here -->
+                    10. "You can't connect the dots looking forward; you can only connect them looking backward. So
+                    you
+                    have to trust that the dots will somehow connect in your future. You have to trust in
+                    something--your gut, destiny, life and karma, whatever. This approach has never let me down, and
+                    it
+                    has made all the difference in my life."
+                    11. "Here's to the crazy ones, the misfits, the rebels, the troublemakers, the round pegs in the
+                    square holes.... The ones who see things differently--they're not fond of rules.... You can
```
LucyMac: This seems a perfect opportunity to use a `<ol>` tag with `<li>`s for each quote.

index.html
```diff
@@ -5,16 +5,158 @@
...
+                    crazy ones, we see genius, because the ones who are crazy enough to think that they can change
+                    the
+                    world are the ones who do."
+                    <button class="button_about_me">
+                        Read More
+                    </button>
+                </div>
+                <div class="col-12 about_me_img">
+                    <img src="product.jpg" class="about_im_2">
+                </div>
```
LucyMac: See my comment in the CSS - this image would be best as a background-image to help it scale better.

css/main.css
```diff
@@ -1,3 +1,203 @@
...
+  border:5px solid gray ;
+  border-radius: 2px;
+} */
+/* I cant make it horizontally*/
+.footer_style{
+  display: flex;
+  flex-direction:column;
+  justify-content: flex-end;
+  align-items: center; 
+  background-color:lightblue;
```
LucyMac: You could move this background-color to the `footer`

index.html
LucyMac: You should avoid have text 'floating' like this without HTML tags around it. 

Also, think of adding a font-family here so it's not the default font.

index.html
```diff
@@ -5,16 +5,158 @@
...
+                        <button class="fix">
+                            More services
+                        </button>
+                    </div>
+                </div>
+            </section>
+    </main>
+    <footer>
+        <div class="fo">
+        <div class="footer_style ">
```
LucyMac: It would be good to set the background-color on the footer, and then on the next level down add a `container` - that your footer content is contained on the page, but the blue colour isn't.

PR #24, opened 2020-02-19
Babak jahangiri - Web Site 2

babakjahangiri:


css/main.css
```diff
@@ -1,3 +1,337 @@
 /* Add your custom CSS here */
 
+@import url('https://fonts.googleapis.com/css?family=Noto+Serif+SC&display=swap');
+@import url('https://fonts.googleapis.com/css?family=Open+Sans&display=swap');
```
LucyMac: Good choice of fonts! 🙂

index.html
```diff
@@ -6,14 +6,151 @@
...
+                    <div class="col-12 sm-col-12 md-col-12 lg-col-12 xl-col-12">
+
+                        <div class="logo-wrapper">
+                            <img src="img/logo_b_t.png" class="logo" alt="B logo" title="B logo">
+                        </div>
+                    </div>
+                </div>
+                <div class="row">
+                    <div class="col-12 sm-col-12 md-col-12 lg-col-12 xl-col-12 center">
+                        <nav>
```
LucyMac: if the number of columns is the same across breakpoints, you don't have to repeat them for each size.
`col-12` works from tiny screen up to infinity - this is the 'mobile-first' approach. So you could just have `col-12` here and it would work the same 😄 
You only need to use the other sizes if you're changing the number of columns on different breakpoints (example mobile 12, then tablet 6 then desktop 4 columns, you would need different classes for each of those, like this: `col-12 md-col-6 lg-col-4`).

LucyMac: Also here you could make that `div` on line 31 the `nav` instead, and that way not have unnecessary wrappers.

index.html
```diff
@@ -6,14 +6,151 @@
...
+        <section class="container" id="project">
+            <div class="row">
+                <div class="col-12 sm-col-12 md-col-12 lg-col-12 xl-col-12">
+                    <p class="part-title">projects.</p>
+                </div>
+            </div>
+            <div class="row">
+                <div class="col-12 sm-col-12 md-col-12 lg-col-4 xl-col-4">
+                    <img class="project-image" src="img/img-projects/p1.jpg" alt="project 1" title="project 1">
+                    <img class="project-image" src="img/img-projects/p2.jpg" alt="project 2" title="project 2">
```
LucyMac: You could wrap these images inside an `<a>` and give them a cool hover effect 💭

css/main.css
```diff
@@ -1,3 +1,337 @@
...
+}
+
+nav > ul > li {
+    list-style-type: none;
+}
+
+nav a {
+    text-decoration: none;
+    cursor: pointer;
+}
```
LucyMac: You could explore adding smooth scrolling for a better effect on click
https://css-tricks.com/almanac/properties/s/scroll-behavior/

babakjahangiri: good suggestion

PR #29, opened 2020-02-20
12 grid project

MahmutSel404:
Hi Lucy,  I have lots of things to fix but in general the work is here. Main problem is the fixation of images sizes. i put columns value in their class section but it looks very different. If you could help me, that would be great. thanks :)


index.html
```diff
@@ -6,15 +6,124 @@
...
+<main>
+<section class="header-hero">
+    
+</section>
+<section class="container hero-content">
+<h3>projects.</h3>
 
-</body>
+<div class="row">
+    <div class="sm-col-12 md-col-12 lg-col-4">
```
LucyMac: You should start your column classes with `col-...` not `sm-col-...` as `sm` starts from 576px, so small mobile phones (like iPhone 4/5) are not included.

index.html
```diff
@@ -6,15 +6,124 @@
...
+    </div>
+</div>
+    
+</section>
+<div class="before-about-img"></div>
+
+<section class="container about">
+    <div class="row hero-content-row">
+        
+        <div class="sm-col-12 md-col-12 lg-col-12">
```
LucyMac: If you're setting the same number of columns for all breakpoints, you don't need to use a class for each one (for example `col-12 sm-col-12 md-col-12 lg-col-12`). You only need a column class for different sizes if you're _changing_ the number of columns for different sizes, for example `col-12 md-col-6 lg-col-4`. 
So in this case just using `col-12` will be enough.

PR #31, opened 2020-02-22
An updated version of the responsive grid project.

byohannes:
An updated version of the responsive grid project.


index.html
```diff
@@ -6,14 +6,232 @@
...
+                          </div>
+                        </div>
+                    
+                    </form>
+                           
+                </div>
+          
+    
+    <!--Footer-->
+    <div class="container section-footer">
```
LucyMac: Use the `footer` tag here instead of a div, and move the footer outside of `main`

index.html
```diff
@@ -6,14 +6,232 @@
...
+        <div class="row">
+
+            <div class=" box col-12 xl-col-3 ">
+                <div >
```
LucyMac: Do you need this extra div?

index.html
```diff
@@ -6,14 +6,232 @@
...
+                    <p> 221 Minard Road</p>
+                    <p> SE6 1NH</p>
+                    <p> Lewisham</p>
+                    <p>London</p>
+                    
+                </div>
+            </div>
+        
+            <div class=" box col-12 xl-col-3">
+                <div>
```
LucyMac: Same: do you need this?

css/main.css
```diff
@@ -1,3 +1,248 @@
...
+  text-align: center;
+}
+.form-container {
+  border-radius: 5px;
+  padding: 20px;
+  color: white;
+}
+@media (min-width: 768px) {
+  .about-col-6 {
+    width: 50%;
```
LucyMac: This is what the `md-col-6` does. Avoid writing your own widths like this - you can use the grid system provided instead.

css/main.css
```diff
@@ -1,3 +1,248 @@
...
+  display: block;
+  font-size: 1.17em;
+}
+
+p {
+  display: block;
+  margin-block-start: 1em;
+  margin-block-end: 1em;
+  margin-inline-start: 0px;
+  margin-inline-end: 0px;
```
LucyMac: This is default browser styling, you don't need this.
What you can do however, is override these default styles with your own, for example by setting margin to `0` for all `p` tags`.

css/main.css
```diff
@@ -1,3 +1,248 @@
...
+input {
+  width: 100%;
+  padding: 12px 20px;
+  margin: 8px 0;
+  display: inline-block;
+  border: 1px solid #ccc;
+  border-radius: 4px;
+  box-sizing: border-box;
+}
+
```
LucyMac: Nice looking inputs 👍

css/main.css
```diff
@@ -1,3 +1,248 @@
...
+}
+h4 {
+  text-align: center;
+  font-weight: bold;
+}
+.row {
+  display: flex;
+  flex-wrap: wrap;
+  margin-right: -15px;
+  margin-left: -15px;
```
LucyMac: Why write this yourself? That's what grid.css is for 🙂 It's all in there already!

css/main.css
```diff
@@ -1,3 +1,248 @@
...
+}
+.aboutus-text {
+  margin: 5rem;
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  align-items: center;
+}
+body {
+  font-size: 20px;
```
LucyMac: Just so you know, the browser adds a default margin of 8px around the entire body. It is common practice to set the body to `margin: 0` to remove this.

css/main.css
```diff
@@ -1,3 +1,248 @@
...
+}
+.about-section {
+  margin: 15px;
+}
+.row2 {
+  display: flex;
+  flex-wrap: wrap;
+  margin-right: 0px;
+  margin-left: 0px;
+}
```
LucyMac: The `row` class (from grid.css) should be used instead here.

index.html
```diff
@@ -6,14 +6,231 @@
...
 
     <title>Responsive grid project</title>
 </head>
 
 <body>
+    
+    <header>
+             <div class="container">
+                <div class="row">
+                    <div class=" top col-12">
```
LucyMac: Do you need to set this to 12 columns, since you're immediately splitting it into 3 and 7 afterwards? I would re-structure it to something like this: 
```
<div class="container">
    <div class="row top">
        <figure class="logo col-5">
            <img src="./images/logo1.png" alt="logo" />
        </figure>

        <nav class="top-navigation col-7">
            <ul>
                //...
            </ul>
        </nav>
    </div>
</div>
```
(note that I have changed the number of columns (5 and 7) so that they add up to 12. The number of columns inside a `row` should always add up to 12)

index.html
```diff
@@ -6,14 +6,231 @@
...
+                         <div class="main">
+                        <h1>Bakerdays</h1>
+                        <p>Personalised Cakes Online and For All Occasions</p>
+                     <button class="shop-now">Shop Now</button>
+                    </div>
+                    
+                </div>
+            </div>
+            </div>
+        </div>
```
LucyMac: Watch your indentation here, it's hard to read.

index.html
```diff
@@ -6,14 +6,231 @@
...
+                    <h5>Price: £35</h5>
+                </div>
+            </div>
+            <div class="col-12 md-col-4">
+                <div class="card">
+                    <img src="./images/events.jpeg" alt="event cakes">
+                    <h5>Events cake</h5>
+                    <h5>Price: £100</h5>
+                </div>
+            </div>
```
LucyMac: Very good split of the columns inside the row here 👍 

Remember to close the `container` and `row` though - you're opening a new container and row on lines 83 and 84, but you haven't closed to the previous ones.

css/main.css
```diff
@@ -1,3 +1,248 @@
...
+  ul {
+    list-style: none;
+    flex-direction: row;
+    justify-content: space-between;
+  }
+  .contact-icons-footer {
+    display: flex;
+    align-items: center;
+    align-content: flex-end;
+    margin-left: 240px;
```
LucyMac: This breaks the page as it makes the element bigger than 100%. Please remove it.

PR #33, opened 2020-02-23
Harris Sheraz

HarrisSheraz20:


css/main.css
```diff
@@ -1,3 +1,133 @@
...
+
+.learn:hover{
+    background-color: gray);
+}
+
+.carRacer{
+    height:700px;
+    width: 670px;
+    margin: auto;
+    float: right;
```
LucyMac: Avoid the `float` rule, it's old and unused. Flexbox should be your go-to for any positioning of this kind 🙂

(in this case, if you follow my comments above about image sizing then you shouldn't need anymore positioning on the image itself).

PR #35, opened 2020-12-20
first commit

alinosratipour:


css/main.css
```diff
@@ -1,3 +1,515 @@
...
+   
+    /* margin-top: 15px; */
+  }
+
+ 
+.card-right-color{
+    background-color: #f4ece6;
+    box-shadow: -4px 12px 21px -4px rgba(0,0,0,0.75);
+    -webkit-box-shadow: -4px 12px 21px -4px rgba(0,0,0,0.75);
+    -moz-box-shadow: -4px 12px 21px -4px rgba(0,0,0,0.75);
```
LucyMac, line 144: This box-shadow should be applied to the entire `projects-box`

resume.html
```diff
@@ -0,0 +1,271 @@
...
+                  <button type="button" class="btn-resume">DOWNLOAD CV</button>
+                  <!-- <button class=" btn btn-primary" >Download CV</button> -->
+             </div>
+         </div> 
+       
+        
+    </div>
+
+
+       <div class="container-fluid my-5">
```
LucyMac, line 94: Remember to add the beige background somewhere here

projects.html
```diff
@@ -0,0 +1,231 @@
...
+                     </div>             
+                    </div>
+                
+                 
+                 <p class="pt-5 project-page-text">I'm a paragraph. Click here to add your own text and edit me. It’s easy. Just click “Edit Text” or double click me to
+                add your own content and make changes to the font. I’m a great place for you to tell a story and let your users know a
+                little more about you.</p>
+               </div>
+               <div class="col-lg-5 col-md-5 col-12 col-sm-12  p-0  ">
+                  <img class="img-fluid  w-100 h-100"  src="https://static.wixstatic.com/media/11062b_6369a82539704f78b4fffed0ef058629~mv2_d_2346_1998_s_2.jpg/v1/crop/x_242,y_0,w_1590,h_1993/fill/w_322,h_404,al_c,q_80,usm_0.66_1.00_0.01/Leaflet.webp">
```
LucyMac, line 108: Inspect what the `image-fluid` class does. And then see the other classes you're adding here. They override each other!

LucyMac: Also remember - don't set a width AND a height on your `img` as this will distort them. `width: 100%` should be enough, and let the height set itself naturally.
In this case, you can just keep `image-fluid` and remove the others.

LucyMac: Or if this doesn't work, set the image as a background-image to better control how it occupied the space.

projects.html
```diff
@@ -0,0 +1,231 @@
...
+           </div>
+         
+        </div>
+
+
+<div class="projects-box mx-lg-auto mx-sm-auto mx-auto ">
+    <div class="row mx-lg-auto my-4">
+        <div class="col-lg-7 col-md-7  col-12 bg-white  card-right-color">
+            <div class="row d-flex flex-lg-row align-items-center  text-start">
+                <div class="col-lg-1  p-0 pt-5 mt-5 blue-box  ">
```
LucyMac, line 119: Watch out with all these Bootstrap utility classes (`mx-auto`, `p-0` etc). You're overriding the grid classes and also losing track of what is happening when. Some of these don't work, others are doing the same thing twice... Remember to use the 'inspect' function on your browser to see what CSS is _actually_ being applied.

LucyMac (commented):
Nice work Ali. Good attempt at copying the design and making a fully responsive site using Bootstrap 👍 Also, well done on your cool mobile nav! Love it. 

There are a few issues with your use of Bootstrap grid classes - don't override what they do with other classes that target the same rules (e.g. using a `col-12` class as well as `p-0`).

You can improve the site still by adding the dark beige background to the pages, and refining the styles of your contact form.

PR #36, opened 2020-12-22
London Class 7  - Tanika - Responsive Grid Project - HTML/CSS - Week 4

Trobinsonshoots:
created the webpage for a therapist, mobile first added first breakpoint still tweaking


index.html
```diff
@@ -4,17 +4,289 @@
...
+            <!-- APPOINTMENT BTN SM DISPLAY -->
+            <div class="display-sm margin-bottom col-10">
+                <button class="appointment-btn">Book an Appointment</button>
+            </div>
 
-    <!-- Add your markup here -->
+            <button class="navbar-toggler p-0 margin-bottom" type="button" data-bs-toggle="collapse" data-bs-target="#navbarSupportedContent" aria-controls="navbarSupportedContent" aria-expanded="false" aria-label="Toggle navigation">
+                <i class="col-1 navbar-toggler-icon fas fa-bars"></i>
+            </button>
+            <div class="display-sm row">
```
LucyMac, line 32: Remember to use a `container` or `container-fluid` class before using `row`

index.html
```diff
@@ -4,17 +4,289 @@
...
+                
+            
+        </nav> 
+    
+    </header>
+    <main class="text-center container">
+        <section class="hero row">
+            <div class="hero-md-img p-0 display-order col-md-6">
+            </div>
+                <img class="hero-img p-0 col-12 display-sm img-fluid" src="./img/erol-ahmed-aIYFR0vbADk-unsplash.jpg" alt="Succulent plant">
```
LucyMac, line 94: You shouldn't have to include the succulent image twice here. Flexbox's `order` rule should be enough? Tip if it's not working, you need to apply the order to both elements, not just the image. 

Also, avoid using `col-...` classes on images directly. Set the col on a div first, and put your image inside.

index.html
```diff
@@ -4,17 +4,289 @@
...
+                        perferendis, saepe soluta culpa voluptatum minus animi eveniet earum. 
+                        Neque reprehenderit sed dolore omnis?Praesentium fuga consequatur 
+                        vitae ut ex et maiores quisquam id voluptates dolorem!</p> 
+                </div>
+                
+            </article>
+            <img class="col-10 offset-1 box-sizing align-self-center" src="./img/pexels-christina-morillo-1181690.jpg" alt="Image of Maria Hope">
+        </section>
+        <section class="therapy-sect row"> 
+            <article class="box-sizing box-bg-1 indiv-therapy col-12 col-md-12 ">
```
LucyMac, line 116: Your text isn't quite vertically aligned. Use `display: flex` with `align-items: center` and equalise your top and bottom padding to help with this.

index.html
```diff
@@ -4,17 +4,289 @@
...
+                        architecto quas atque.</p>
+                </div>
+                
+            </article>
+        </section>
+        
+        <section class="article-section row justify-content-center">
+            <h2 class="article-heading">Articles about Psychology</h2>
+            <!-- Articles carousel SM screen -->
+            <div id="carouselExampleSlidesOnly" class="display-sm article-carousel carousel slide col-11" data-bs-ride="carousel" data-interval="touch">
```
LucyMac, line 150: Nice carousel 🎠

css/about.css
```diff
@@ -0,0 +1,202 @@
...
+    box-sizing: border-box;
+    font-family: 'Quicksand', sans-serif;
+    color: var(--text-color);
+    margin: 0;
+    padding: 0;
+    overflow-X: hidden;
+}
+
+header {
+    padding-bottom: 20px;
```
LucyMac, line 26: This isn't getting applied because you've got `p-0` on your header. I don't think you need `p-0`

css/blog.css
```diff
@@ -0,0 +1,238 @@
...
+    padding-right: 27px;
+}
+
+.chat-icon-fixed-padding {
+    padding: 20px 40px;;
+}
+/* FOOTER */
+
+.footer-section {
+    padding: 60px 0 80px 0;
```
LucyMac, line 180: This is breaking your layout (it causes the white space on the right-hand side) because you're overriding the left- and right padding of the container. Only set the top and bottom padding here.

css/contact.css
```diff
@@ -0,0 +1,188 @@
...
+}
+
+.chat-icon-fixed-padding {
+    padding: 20px 40px;
+}
+
+/* FOOTER */
+
+.footer-section {
+    padding: 60px 0 80px 0;
```
LucyMac, line 139: Same issue here as for blog.css

LucyMac (commented):
Awesome work Tanika. Very well organised code, good naming, excellent replicate of the original design, and good use of Bootstrap layout classes - just remember to always use a `container` before `row`, and to avoid overriding any side padding or margin on Bootstrap elements!
You could add a few more responsive rules to reduce your heading font sizes and section spacing on mobile, but otherwise really well done 🌟

css/contact.css
Trobinsonshoots: awesome, will change both

index.html
Trobinsonshoots: 100% agree don't know why I did that must of been rushing or it did register at the time. adding 'col..' to images isn't advised, I used it because it place it perfectly for how I wanted it to be displayed and to achieve that I may of had to use 'object-fit', which again not advised because of IE not being compatible. What would be the best practice for placing images with bootstrap? Or is simply adding 'col..' to the div going to achieve that? i will attempt it myself but if there is a better way of doing so that would be great to know and add to my arsenal.

index.html
Trobinsonshoots: will do, think there may of been an issue why I didn't, but with it being common practice I should have done it and fixed the issue rather than leave it out.

css/blog.css
```diff
@@ -0,0 +1,238 @@
...
+
+header {
+    padding-bottom: 20px;
+}
+
+h1, h3 {
+    font-family: 'Libre Baskerville', serif;
+    font-style: italic;
+    font-size: 1em;
+    line-height: 1.7em;
```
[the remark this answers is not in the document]
Trobinsonshoots: awesome, will start implementing that. Will make css files less cumbersome and easier to read.

PR #37, opened 2020-12-23
Adding css.

rfloresram:


index.html
```diff
@@ -7,14 +7,150 @@
...
+        </nav>
+    </header>
+
+<!-- Main content -->
+
+    <main class="primary-content">
+        <section class="hero-main">
+            <img class="col-12" src="./img/suculent.webp" alt="Succulent plant">
+
+            <section class="mint-container col-10 offset-1 col-md-8 offset-md-2">
```
LucyMac, line 68: Follow the Bootstrap rules: you're using `col-...` classes here without a `container` or a `row`.

Try applying `container-fluid` to the top level (either once on the `main`, or on each section individually). 
Then, always use `row` before you use `col-...`

index.html
```diff
@@ -7,14 +7,150 @@
...
+                <h2>Online Therapy for Emotional Health</h2>
+                <p>I am a description. Click here to edit.</p> 
+            </section>
+        </section>
+
+        <section class="about-maria">
+            <article class="about-psychologist">
+                <div class="col-md-8 offset-md-2" >
+                    <h3 class="col-12 text-center">About</h3>
```
LucyMac, line 77: You cannot use `col-12` directly after another `col-...` element. 
Remove `col-12` here you don't need it. Same for the next 2 elements.

LucyMac: Again, you need to use `row` on the article as it is directly before your `col-` class, and you need a `container` or `container-fluid` before that (this can be on the section, or on the `main`, but not both).

index.html
```diff
@@ -7,14 +7,150 @@
...
+            </article>
+
+ <!-- Articles boxes sections -->
+ </article>
+        
+        <section>
+            <img class="col-10 maria-maria align-self-center" src="./img/psychologist_woman.webp" alt="Maria Hope Psychologist">
+        </section>
+
+        <section class="services-section">
```
LucyMac, line 90: Add a `row` here.
It will fix your layout on desktop because `row` has `display: flex`

index.html
```diff
@@ -7,14 +7,150 @@
...
+ <!-- Articles boxes sections -->
+ </article>
+        
+        <section>
+            <img class="col-10 maria-maria align-self-center" src="./img/psychologist_woman.webp" alt="Maria Hope Psychologist">
+        </section>
+
+        <section class="services-section"> 
+            <article class="text-w-box box-n-1 col-md-4 ">
+                <div class=" offset-1 col-md-8 offset-md-2">
```
LucyMac, line 92: Again, don't use `col-...` classes on both these elements. Only on the one directly after `row`.

LucyMac (commented):
Good start Rafael. 
Please review how Bootstrap works: you must wrap each section inside a container, and then remember to always use `row` before you use any `col-...` class

Can you also add one or two more pages to allow navigation to the pages in the header nav? The contact page would be a good CSS exercise.

PR #39, opened 2020-12-23
first draft of home page

nPelvin:
First draft of home page, only landing page no links working yet.


index.html
```diff
@@ -4,16 +4,196 @@
...
+            <a class="socials" href="#"><img src="img/iconmonstr-twitter-1.svg" alt="Twitter"></a>
+            <a class="socials" href="#"><img src="img/iconmonstr-linkedin-1.svg" alt="LinkedIn"></a>
+            <a class="socials" href="#"><img src="img/iconmonstr-instagram-11.svg" alt="instagram"></a>
+
+        </div>
+    
+</section>
+</div>
+
+        <div class="col-md-5">
```
LucyMac, line 91: On mobile this whole text section looks like it should not span the whole width, but perhaps be 10 columns, with an offset of 1.

css/projects.css
```diff
@@ -0,0 +1,264 @@
...
+
+/* CSS for Profile Section */
+.jobposition{
+    font-weight:700;
+}
+.date{
+    color:#0150FD;
+    padding-top:3em;
+    font-weight:700;
+}
```
LucyMac, line 79: Could you add something here to re-create the vertical blue bar from the design? 🙂

css/projects.css
```diff
@@ -0,0 +1,264 @@
...
+    padding: 3px 16px;
+    border:solid black 2px;
+    font-weight:600;
+    font-size:1.3em;
+    margin:10px 0px;
+}
+.capture{
+    width:100%;
+    height:100%;
+}
```
LucyMac, line 155: Don't set a width AND a height on your `img`s as this will distort them. `width: 100%` should be enough, and let the height set itself naturally.

To better control the width of the image:
- wrap it inside a `div`
- give that `div` a width (and an optional height)
- and then just `width: 100%` on the `img` (no height)

Of course this can only work if your images are already the right aspect ratio. In this case, they probably aren't, so you could consider setting them as background images instead, which is easier to make fill the space using `background-size: cover`

Up to you! 🙂

LucyMac (commented):
Nice work Nigel. Good replica and nice use of Bootstrap. 
See a few comments to avoid common pitfalls with Bootstrap's grid system (in short, don't override it) and also on sizing images. 
Could you add some custom fonts as well?

PR #40, opened 2020-12-25
London Class 7 - Askin Ekinci - Responsive Grid Project - HTML/CSS - Week 4

askinekincihf:


clairesmi:
Hi Askin, this looks great! Your layout is really nice and the responsiveness works really well. I really like your contact page, the form especially looks good :)

In your CSS file, take care to place your imports at the top of the document rather than part of the way down, to make the file more easily readable both for yourself and others. Good, descriptive class names - keep this up, the same goes for your semantic html tags!

You had a slight issue with the format of your pathnames for both your images and `href` attributes in your anchor tags. When you write these, try to think of moving from the file that you're currently writing in, to the file that you want to reference. So from your `index.html` file to get to your `about.html` file you would need to move up one level which is signified by `.` and then down into the html folder to the `about.html` file, which is signified by `/` . So where you have written `<a href="/html/about.html">` it should be `<a href="./html/about.html">` - try this and it should sort your broken images and links. 

Great job overall, I hope you enjoy the JavaScript module!

askinekincihf:
Hi Claire, thank you very much for the feedback. It is very much appreciated. 
I have modified the pathways of the hrefs and img links.  Also, I moved the import to the top of CSS.

PR #41, opened 2020-12-28
London class 7-hozan ali-responsive-grid-project-week4-HTML/CSS

Hozan94:


index.html
```diff
@@ -6,14 +6,324 @@
...
     <title>Responsive grid project</title>
 </head>
 
 <body>
+    
+                        <!-- Adding a nav bar and a toggler -->
+<header>
+    <section class="fixed-top bg-light">
+        <aside class="d-none d-lg-flex justify-content-end mx-5 pt-3">
+            <a class="booking" href="#">Book an Appointment</a>
```
louisechow, line 31: I think I would have chosen to use the `button` tag here. Looking at the original website, this certainly looks like it's a button, and there's an expectation that when I click on it. But what you have here works too.

One thing I did notice is that in the last section of the page there is a button that looks exactly like this one but it has a different class (this is called booking and the other is called view-article). 

You could reduce repetition if you have just one class that had the button styling and then you reused it on the different elements. That way you don't repeat any code and if you ever need to update the style of a button on your page, you only need to do it in one place.

index.html
```diff
@@ -6,14 +6,324 @@
...
+                <h5>03.</h5>
+                <a href="#">Couples Therapy</a>
+                <p>Lorem ipsum, dolor sit amet consectetur adipisicing elit. Labore obcaecati possimus sit iusto nobis ad dolore repudiandae doloribus culpa facilis!</p>
+            </article>
+        
+        </article>
+    </section>
+                       
+                        <!-- Adding a Div to act as a space between two sections -->
+    <div class="space" style="height: 280px;"></div>
```
louisechow, line 124: I think you should try to use CSS to add whitespace between the sections instead of adding this `div` in. Remember, you want to try and have all your style related code in CSS, not in the HTML.

index.html
```diff
@@ -6,14 +6,324 @@
...
+            
+            </div>
+                
+        </nav>
+    
+    </section>
+</header>
+                       
+<main>                  <!-- Adding top part of the page (a picture and "Online Therapy for Emotional Health" part) -->
+    <article class="top-page mx-lg-5">
```
louisechow, line 61: I would try to avoid putting words that describe the layout in the name of the class - e.g. **top**, **middle**, **bottom**. Leave layout specifications for the code in your CSS blocks.

The reason being, imagine i'm Maria who you've built this website for and now decide I want to change the order of the sections on the page. Not only will you need to rearrange the sections in `index.html` but you'll need to change all your class names too because they are all inaccurate now.

I'd suggest giving your sections names that correspond to the type on content in them. The items in the navigation bar give a good indication of what each of the sections is about; about, services, blog.

index.html
```diff
@@ -6,14 +6,324 @@
...
+        <h5 class="text-center">Articles about Psychology</h5>
+        
+        <div id="blogsCarousel" class="carousel" >
+            <ol class="carousel-indicators">
+                <li data-target="#blogsCarousel" data-slide-to="0" class="active"></li>
+                <li data-target="#blogsCarousel" data-slide-to="1"></li>
+                <li data-target="#blogsCarousel" data-slide-to="2"></li>
+            </ol>
+            
+            <div class="container">
```
louisechow, line 217: This was the only container definition that I found in this file. What container to rows in the above section belong to?

LucyMac (commented):
Nice work, your site looks great! 
Watch out though, using all those Bootstrap spacing classes alongside the grid classes causes code clashes and can misalign the grid system. I've highlighted some examples below.

index.html
Hozan94: You are right, it does not belong to any rows, the only reason I added this is because it added a space around the carousel.
Otherwise, when removed you can see the carousel is nearly touching the edges of the screen (when In Mobile view)

index.html
Hozan94: Thank you for this suggestion.

index.html
Hozan94: thanks for your insight, I will follow this rule from now on.

index.html
Hozan94: Hi Louise, I have different class names for each button as each one of them had few different stlyes such as size and fonts so they look a bit different from each other.
Hence, the reason not knowing how to have a same class name for all the buttons . But overlall, I agree with your feedback.

contact.html
```diff
@@ -0,0 +1,152 @@
...
+
+    <!-- Importing Google fonts  -->
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Libre+Baskerville&display=swap" rel="stylesheet">
+
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Libre+Baskerville:ital@0;1&display=swap" rel="stylesheet">
+
+    <link rel="preconnect" href="https://fonts.gstatic.com">
+    <link href="https://fonts.googleapis.com/css2?family=Quicksand:wght@300&display=swap" rel="stylesheet">
```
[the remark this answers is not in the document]
Hozan94: @LucyMac  thank you for this info. 

> Nice work, your site looks great!
> Watch out though, using all those Bootstrap spacing classes alongside the grid classes causes code clashes and can misalign the grid system. I've highlighted some examples below.

@LucyMac Many thanks for the feedback, all the suggestions you mentioned are noted and will try to follow them.

contact.html
```diff
@@ -0,0 +1,152 @@
...
+    
+                </div>
+    
+            </nav>
+    
+        </section>
+    </header>
+    
+    <main>
+        <Section class="contact-page pt-5">
```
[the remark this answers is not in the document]
Hozan94: @LucyMac  I did not add a `container` as I wanted to the design the bleed to the edges of the page, but you are right `conatiner-fluid` would have done that.

contact.html
```diff
@@ -0,0 +1,152 @@
...
+            <article class="my-contact row g-0">
+                <h1>Contact Me</h1>
+                <h5>I'm a title. ​Click here to edit me.</h5>
```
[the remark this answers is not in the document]
Hozan94: @LucyMac  Thanks for this suggestion.

contact.html
```diff
@@ -0,0 +1,152 @@
...
+                <p class="col-lg-4 offset-lg-4 pt-1 pb-lg-4">I'm a paragraph. Click here to add your own text and edit me. I’m a great place for you to tell a story and let your
+                users know a little more about you.</p>
+            </article>
+       
+            <form action="" target="_blank" class="container my-forms" >
+                <input class="col-12 col-lg-3 offset-lg-3 mb-2 p-2 " type="text" name="name" placeholder="Name" title="Name" required >
+                <input class="col-12 col-lg-3 mb-2 p-2" type="email" name="email" placeholder="Email" title="email" required >
+                <input class="col-12 col-lg-6 offset-lg-3 mb-2 p-2" type="text" name="subject" placeholder="Subject" title="subject" required>
```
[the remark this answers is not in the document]
Hozan94: @LucyMac you are right, you mentioned this in class as well, and they only reason I used padding/margin is so I have wide input buttons, when removed they become narrow. So if there is a way around this without clashing with the default paddings/margins then that would be appreciated.

index.html
```diff
@@ -6,14 +6,324 @@
...
+                        <path fill-rule="evenodd" d="M8 1a.5.5 0 0 1 .5.5v11.793l3.146-3.147a.5.5 0 0 1 .708.708l-4 4a.5.5 0 0 1-.708 0l-4-4a.5.5 0 0 1 .708-.708L7.5 13.293V1.5A.5.5 0 0 1 8 1z" />
+                    </svg> 
+                </a>
+            </article>
+        </section>
+    </article>
+                        
+                        <!-- Adding the middle part of the page (About Maria Hope and a picture) -->
+    <section class="middle-page mt-5 pt-5 mx-lg-5 text-center">
+        <article class="row g-0">
```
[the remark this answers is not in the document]
Hozan94: @LucyMac I tried adding a `container` class, it changed the width even after removing the `mx-lg-5` , is there a way to achieve this and still have its original layout?

PR #42, opened 2020-12-28
London Class 7 - Domenico Delli Carri - Responsive Grid Project - HTML/CSS - Week 4

Domenico-cyf:
London Class 7 - Domenico Delli Carri - Responsive Grid Project - HTML/CSS - Week 4


css/main.css
```diff
@@ -1,3 +1,58 @@
...
+    ;
+}
+
+.flex-nav{
+    display: flex;
+    justify-content: space-between;
+    padding: 5vh;
+}
+h1 h2 h3 h4 h5 h6 {
+    font-family: "https://fonts.google.com/specimen/Poppins";
```
LucyMac, line 24: Again here you need to say `font-family: 'Poppins', sans-serif;` not include the link. The link to the asset should be inserted in the `head` of your HTML. 
See the right-hand panel [here for instructions](https://fonts.google.com/specimen/Poppins?query=poppins&selection.family=Poppins:wght@300&sidebar.open=true)

css/main.css
```diff
@@ -1,3 +1,58 @@
...
+}
+.background-beige{
+    background-color: #E6DACD;
+    margin: 0px;
```
LucyMac, line 28: Problem here: you are using this class on the same level as Bootstrap classes `col-md-5 col-12`. 
Whenever you apply your own custom class to an element that also has Boostrap grid classes, you **must not** include any left or right margin, padding or width, as it will interfer with how Bootstrap works. 
You can however keep using this class for the `background-color` and `height`, just remove the margin rule.

css/main.css
```diff
@@ -1,3 +1,58 @@
...
+    height: 50vh;
+}
+.container-class{
+    margin: 0px;
+}
```
LucyMac, line 33: Again, don't set your own margins on a class that is applied to a Bootstrap element (this is used on a `container` element)

index.html
```diff
@@ -7,13 +7,62 @@
...
+                <a href="">RESUME</a>
+                <a href="./projects.html">PROJECTS</a>
+                <a href="./contact.html">CONTACT</a>
+            </nav>
+        </section>
+    </header>
+
+
+    <main>
+        <div class="container container-class">
```
LucyMac, line 29: Remove your custom class here as its margin interfers with the Bootstrap container class

index.html
```diff
@@ -7,13 +7,62 @@
...
+            <div class="row">
+                <div class="col-md-5 col-12 background-beige">
```
LucyMac, line 31: See my comment about the `background-beige` class (it shouldn't include a margin) but otherwise this is a good use of Boostrap classes. 
You've got a `container`, then a `row`, then `col-....` classes so all good here 👍

index2.html
```diff
@@ -0,0 +1,97 @@
+<!DOCTYPE html>
```
LucyMac, line 1: A note on the name of this HTML file. Typically, `index` is the name we give to the homepage. The other pages can be named after their content, for example `resume` or `projects` or `about-me`

index2.html
```diff
@@ -0,0 +1,97 @@
...
+          </nav>
+          
+    </header>
+
+
+    <main>
+        <div class="container">
+            
+            <div class="row row-text">
+                <div class="offset-1 col-10 offset-1">
```
LucyMac, line 41: You don't need to say `offset-1` again here, it will leave a gap automatically.

index2.html
Domenico-cyf: In reality, those are tests and not quite sure which version will be the index, so, for now I have 2 versions of the index page.

index2.html
Domenico-cyf: the offset is to center the element which takes 10 columns instead of having it at the left hand side

Domenico-cyf:
Probably the whole index page will be disregarded and start afresh since I don't see an easy way out.

NewTest.html
```diff
@@ -0,0 +1,49 @@
...
+    border: 3px solid green;
+}
+.my-row{
+    border: 3px solid red;
+    height: 300px;
+}
+.my-col{
+    border: 3px dotted blue;
+}
+</style>
```
LucyMac, line 25: Don't include CSS in your HTML files, it's best practice to have them in separate files. I realise this might just be you experimenting 😉

resume_pj/css/mainR.css
```diff
@@ -15,21 +14,23 @@ a {
...
+.hello-section {
+    margin-top: -100px;
+    padding-top: 120px;
+    background-color:  #E6DACD ;
 }
-h1 h2 h3 h4 h5 h6 {
+
+h1, h2, h3, h4, h5, h6 {
     font-family: "https://fonts.google.com/specimen/Poppins";
 }
```
LucyMac: You can't put the link to a font directly in your CSS. This should be `font-family: 'Poppins', sans-serif;` and the link to t he font should be imported into your HTML file in the `head` - see instructions on the right panel here: https://fonts.google.com/specimen/Poppins?sidebar.open=true&selection.family=Poppins:wght@200

resume_pj/index.html
```diff
@@ -0,0 +1,148 @@
...
+            </div>
+                 <div class="border border-danger col-md-7 col-12 hello-section">
+                    <h1>Hello</h1>
+                    <h3>Here's who I am & what I do</h3>
+                    <p>
+                        I'm a paragraph. Click here to add your own text and edit me. It’s easy. Just click “Edit Text” or double click me to add your own content and make changes to the font.
+
+ 
+
+I’m a great place for you to tell a story and let your users know a little more about you.
```
LucyMac, line 93: BEar in mind TML doesn't take line breaks like this into account. If you want to split the 2 parts of the paragraph, use separate `p` tags for each

resume_pj/index.html
```diff
@@ -0,0 +1,148 @@
...
+            <div class="col-6">
+               <h3>Write</h3>
+               <h4>123-456-7890</h4>
+                
+            </div>
+            <div class="col-12">
+                <h3>Follow</h3>
+                <div class="row social-class justify-content-center">
+                    <div class="col-1">
+                        <a href="">
```
LucyMac, line 120: Correct use of Bootstrap here re-using `row` when nesting columns inside columns. I'm not sure this was necessary though. You can just take this whole section from line 116 to 140 and take it outside of the `row` from line 102, and put in inside its own `row`, like this:

```
<footer class="container-fluid">
     <div class="row">
            <div class="col-6">
                <h3>Call</h3>
                <h4>123-456-7890</h4>
            </div>
            <div class="col-6">
               <h3>Write</h3>
               <h4>123-456-7890</h4>
            </div>
      </div> // close the first row
      <div class="row> // open a new one
            <div class="col-12">
                <h3>Follow</h3>
            </div> // give the 'follow' heading its own 12 columns if you always want it to occupy the full width
            <div class="col-12 social-class justify-content-center"> // no need for a new row here
                    <div class="icon-wrapper"> // I don't think you need to use col-1 here, just give them a custom class like 'icon-wrapper' with fixed width (e.g. 50px)
                        <a href="">
                            <i class="fa fa-instagram"></i>
                        </a>
                    </div>
                    <div class="icon-wrapper">
                        <a href="">
                            <i class="fa fa-facebook"></i>
                        </a>
                    </div>
                    (etc for other icons)
```

resume_pj/index.html
```diff
@@ -0,0 +1,148 @@
...
+                    <img src="./img/photo-profile.jpeg" class="rounded-circle col-5" alt="profile pic">
+                    <h2>Maya Nelson</h2>
+                    <hr>
+                   
+                        <div class="row justify-content-around">
+                            
+                                <button type="button" class="col-5 btn btn-outline-primary">RESUME</button>
+                            
+                            
+                                <button type="button" class="col-5 btn btn-outline-primary">PROJECTS</button>
```
LucyMac, line 56: Don't use Boostrap buttons please. Use Bootstrap just for the layout / responsive behaviour. Your buttons and everything else should be styled by you with custom CSS.

LucyMac (commented):
Good use of Bootstrap overall, but you need to add more of your custom CSS.

A few comments below.

PR #43, opened 2020-12-28
Updated Pages

jmodiale:


css/projects.css
```diff
@@ -0,0 +1,59 @@
...
+
+a {
+    padding: 15px;
+    text-decoration: none;
+    color: black;
+}
+
+.header-text {
+    font-size: 25px;
+    margin-left: -80px;
```
LucyMac, line 21: You're not using the grid system correctly here. Don't give the blue square its own column class - just put it inside the same `col-7` as the header text, so they can sit next to each more easily.

css/resume.css
```diff
@@ -0,0 +1,76 @@
...
+
+.container header {
+    padding-top: 40px;
+}
+
+.col-md-2 {
+    text-align: center;
+}
+
+#btn {
```
LucyMac, line 61: Why are you using an ID here?
Don't use IDs in your CSS. You can use IDs for JavaScript, but when adding styles to HTML, target your elements using classes.

index.html
```diff
@@ -1,20 +1,101 @@
...
 
 <body>
+    <section class="container">
+        <header class="row">
+            <div class ="col-sm-1" >
+                <img class="blue" src="./img/blue.jpg">
+            </div>
+            <div class="col-7">
+                <p class="header-text"><span>Maya Nelson</span> / Project Manager</p>
+            </div>
```
LucyMac, line 28: combine these so the blue square is inside the same `col-` class as the header-text

index.html
```diff
@@ -1,20 +1,101 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Dancing+Script:wght@400;500&family=Euphoria+Script&family=Redressed&family=Satisfy&display=swap"
+    rel="stylesheet">
+    <link rel="stylesheet" href="css/main.css">
 
     <title>Responsive grid project</title>
 </head>
 
 <body>
+    <section class="container">
+        <header class="row">
```
LucyMac, line 22: `header`, `main` and `footer` should be the first level of separation. Split your page into those 3 blocks, and _then_ add sections inside.

index.html
```diff
@@ -1,20 +1,101 @@
...
+            <div class ="col-sm-1" >
+                <img class="blue" src="./img/blue.jpg">
+            </div>
+            <div class="col-7">
+                <p class="header-text"><span>Maya Nelson</span> / Project Manager</p>
+            </div>
+            <div class="col-md-4">
```
LucyMac, line 29: 4 column isn't quite wide enough for your nav to fit in one line below 1200px width. How can you fix this using Bootstrap's responsive gris system?

css/main.css
```diff
@@ -1,3 +1,152 @@
...
+
+.card img {
+    padding-top: 10px;
+    align-self: center;
+    padding-bottom: 10px;
+}
+
+#idcard {
+    margin-left: 280px;
+}
```
LucyMac, line 89: Don't use IDs

LucyMac: This `margin-left` defeats the purpose of using Bootstrap for your layout. Can you use an `offset` class instead?
You also need to account for mobile - this margin-left is making that card disappear because there is no room for it.

css/main.css
```diff
@@ -1,3 +1,152 @@
...
+
+.card {
+    height: 500px;
+    background-color:rgb(238, 237, 243);
+}
+
+aside {
+    margin-right: 250px;
+}
```
LucyMac, line 98: Same here - this margin-right is not necessary: use Bootstrap to create the space you need. 
It looks like you want this entire `main` section to use up 8 columns instead of the full width. So wrap it inside `col-md-8` and use `offset-md-2` to center it.

LucyMac (commented):
Joy, you make a good attempt at using Bootstrap classes properly, however you must avoid over-riding these with your own margins - see comments below. 
You also have not accounted for the layout on mobile or used any media queries. It's best practice to code 'mobile-first', meaning you code everything to look good on mobile by default, and then you use Bootstrap's responsive classes + your own media queries to make it work on desktop as well. 
This task was to re-create the exact design provided which is a required skill for front-end developers: the client usually approves a specific design and it's your job to re-create it perfectly using the same fonts, colour etc. 
Can you make changes to this please so it (1) works on mobile and (2) follows the design more closely?

PR #44, opened 2020-12-29
London Class-7-Yasemin Bayraktutan-Responsive Grid Project-HTML/CSS Week 4

ybayraktutan:


css/main.css
```diff
@@ -1,3 +1,96 @@
...
+.services {
+    min-height: 500px;
+}
+
+.more-services {
+    margin-top: -150px;
+}
+
+.myDIV {
+    transform:translateY(-200px);
```
LucyMac, line 82: Good work achieving the overlapping image effect 👏 
Try and keep your class name casing consistent - it's common practice to keep CSS class names lower case and separate words with dashes `-`. Also try to use more descriptive names 😉

css/main.css
```diff
@@ -1,3 +1,96 @@
...
+}
+
+@media (max-width: 600px) {
+    .more-services {
+        margin-top: 0px;
+    }
+    .logo {
+        text-align: center;
+    }
+  }
```
LucyMac, line 96: This is a desktop-first approach, meaning your code works for desktop by default, and then your media queries target the mobile size (from `0px` until `600px`). 
It's best practice nowadays to follow a mobile-first approach: you write your code for mobile only, and then add media queries (using **min**-width) to make it work on desktop.

index.html
```diff
@@ -6,15 +6,173 @@
...
+        <!-- Hero Section -->
+        <section class="container-fluid my-5">
+            <div class="row my-5 text-center hero">
+                <div class="col-12 col-sm-6 order-sm-2 p-0 ">
+                    <img class="img-fluid w-100 h-100" src="./img/first image.webp" alt="Hero Section Image">
+                </div>
+                <div class="col-12 col-sm-6 order-sm-1 bg-lightmint d-flex flex-column justify-content-center ">
+                    <h1 class="fw-lighter lh-base my-3">
+                        Online Therapy for Emotional Health
+                    </h1>
```
LucyMac, line 73: Don't rely too much on Bootstrap for aesthetics, fonts, sizing, spacing etc. Focus on using Bootstrap for its grid system only (containers, rows and columns) and write your own CSS for everything else. 

Here for example, you could set the font-size to be bigger on mobile by default (34px in the design), and then grow it to 49px on desktop using a media query in your CSS.

index.html
```diff
@@ -6,15 +6,173 @@
...
+        <section class="container-fluid ">
+            <div class="row about py-5 bg-lightgrey justify-content-center ">
+                <div class="col-10 col-md-6 col-lg-4 text-center h-100">
+                    <h6 class=" fst-italic fw-lighter pt-5">About</h6>
+                    <h1 class=" fw-lighter lh-base">Maria Hope</h1>
+                    <p class="pt-3">I'm a paragraph. Click here to add your own text and edit me. It’s easy. Just click “Edit Text” or double click me to add your own content and make changes to the font. Feel free to drag and drop me anywhere you like on your page. I’m a great place for you to tell a story and let your users know a little more about you.</p>
+                    <a class="mt-3" href="#"><img src="./img/arrow-icon.svg" alt="arrow-icon"></a>
+                </div>
+            </div>
+            <div class="myDIV mt-n5" >
```
LucyMac, line 91: `mt-n5` doesn't seem to be doing anything.

css/main.css
LucyMac: If you want to better control the white spacing below the image, try setting `position: absolute` to it (this will take it out of the flow of the page so the other elements will act as if it wasn't there). Then add some `padding-bottom` to the section so it forces some space for it.

index.html
```diff
@@ -6,15 +6,173 @@
...
 </head>
 
 <body>
+    <!-- Header -->
+    <header class="container">
+        <div class="row">
+            <nav class="navbar navbar-expand-sm bg-white navbar-light fixed-top">
+               
+                <div class="col-12 col-sm-8 order-sm-2 justify-content-around d-flex flex-sm-column align-items-sm-end"> 
+                    <a class="col-8 col-sm-5 col-md-4 col-lg-2 appointment-button bg-lightmint text-decoration-none text-center" href="#">Book an Appointment</a>
```
LucyMac, line 28: Dont' use columns inside of columns without using a new `row` in between them as it will mess with the alignment of elements.

contact.html
```diff
@@ -0,0 +1,165 @@
...
+                </div>
+                <div class="col-md-6 mb-2 mx-3">
+                    <input type="text" class="form-control" placeholder="Subject" id="usr">
+                </div>
+                <div class="col-md-6  mb-2 mx-3">
+                    <textarea class="form-control" rows="5" id="comment">Comment</textarea>
+                </div>
+                <div class="col-md-6 mb-2 mx-3">
+                    <button class="appointment-button bg-mediummint my-2 w-100">Send</button>
+                </div>
```
LucyMac, line 88: Your form looks great - could yo try and follow the design a bit more precisely though (black box on hover and focus) instead of relying on Bootstrap? As a front-end dev, you would be expected to follow the design precisely 😉 
Try the pseudo element `:focus`

LucyMac (commented):
Good work Yasemin, your site looks good on small and large devices 👍 
Try and code in a mobile-first way though (see my comments below), and also avoid relying on Bootstrap for your styling - use it mostly for its grid.

ybayraktutan:
> Good work Yasemin, your site looks good on small and large devices 👍
> Try and code in a mobile-first way though (see my comments below), and also avoid relying on Bootstrap for your styling - use it mostly for its grid.

PR #45, opened 2020-12-29
MyInitialCommit

AhmadJanah:


css/main.css
```diff
@@ -1,3 +1,373 @@
...
+}
+
+.photo-div {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+}
+
+.my-photo {
+  width: 300px;
```
LucyMac, line 102: Can you adapt the size of your photo for smaller screens? It's currently too big between 570px and 1000px screen width

css/main.css
```diff
@@ -1,3 +1,373 @@
...
+  width: 100%;
+  margin-bottom: 0px;
+  margin-top: 45px;
+}
+
+.small-photo {
+  width: 25px;
+}
+
+.Intro-div {
```
LucyMac, line 128: It's common practice to use only lower case letters in class names: `intro-div`
(also you can just call this `intro` so it's not tied to the HTML element you are using...)

css/main.css
```diff
@@ -1,3 +1,373 @@
...
+  font-family: Verdana, Geneva, Tahoma, sans-serif;
+  width: 150px;
+  border-radius: 15px;
+  margin-left: 20px;
+}
+
+.res-btn:hover {
+  background-color: white;
+  color: rgb(58, 76, 240);
+}
```
LucyMac, line 177: Very neat 👍 
Avoid using `width` for buttons in case the text inside changes - you want your button to adapt to its content. Instead, just add `padding` and the button will change size according to the text.

css/main.css
```diff
@@ -1,3 +1,373 @@
...
+
+.smaler-photo {
+  width: 20px;
+}
+
+.footer-small-row {
+  display: none;
+}
+
+@media (max-width: 1140px) {
```
LucyMac, line 222: We want to be coding in a "mobile-first" way: this means you should write all your code to look good on mobile by default, and _then_ you add media queries (using **min**-width) to adapt it for desktop.

css/resume.css
```diff
@@ -0,0 +1,70 @@
...
+  justify-content: center;
+}
+
+.normal-row {
+  height: 100px;
+}
+
+.narrow-row {
+  height: 50px;
+}
```
LucyMac, line 13: You want your content to dictate how high elements are, not give each row a fixed height.
You can do all this with top and bottom margin and padding: you don't need separate rows for each separation between the white blocks, just add `margin-bottom` to each white block. 

Your blocks themselves should use `col-md-8 offset-md-2` to give them their width and center them. 

Then, inside each white block, you only need **1** row, and the content inside can be split by using `col-12` whenever you want to use the whole width.

resume.html
```diff
@@ -0,0 +1,515 @@
...
+                <div class="col-4">
+            
+                </div>
+                <div class="col-4">
+            
+                </div>
+                <div class="col-2 gray-div">
+            
+                </div>
+            </div>
```
LucyMac, line 151: This is not the correct way to create empty space with Bootstrap: see my other comment about about using Bootstrap's `offset` classes and also [the docs](https://getbootstrap.com/docs/4.0/layout/grid/#offsetting-columns)

resume.html
```diff
@@ -0,0 +1,515 @@
...
+            <div class="row">
+                <div class="col-2 gray-div">
+            
+                </div>
```
LucyMac, line 302: All these divs just to create space 😱 You don't need to do this.
You should only use HTML elements to wrap content (apart from a few exceptions). Bootstrap offers everything you need to create blank space ([offset](https://getbootstrap.com/docs/4.0/layout/grid/#offsetting-columns)), and also you don't need to have a separate `row` for everything. You can have several `col-12` blocks inside one `row` and they will just sit on top of each other 🙂

resume.html
```diff
@@ -0,0 +1,515 @@
...
+                <div class="col-8">
+                   <h3> Professional Skill Set</h3>
+                </div>
```
LucyMac, line 305: This whole section should just be a list:
```
                <ul>
                    <li>Entrepreurial mindset</li>
                    <li>Critical thinking</li>
                    <li>Attention to detail</li>
                    <li>Teamwork & collaboration</li>
                </ul>
```
With added CSS classes to make it match the design.

LucyMac: (you want these to be all stacked on mobile, and spread horizontally on desktop)

LucyMac (commented):
Good Ahmad - be careful how you use Bootstrap, see my comments below about using `offset` to create space, and limiting how many rows you actually need. 
Could you improve your resume section to not use all the empty divs and instead limit your HTML to only what you need for your content?
I think a little more top and bottom padding inside your different blocks, as well as importing and applying the project fonts will make it look even better 😉

resume.html
```diff
@@ -0,0 +1,515 @@
...
+                    <p>UAE - Abu Dhabi</p>
+                </div>
+                <div class="col-4">
+                    <p> Full ERP technical support for a variety of complex projects.</p>
+                    <p> Managed and coordinated team priorities with project managers, functional consultants, and end-users, and achieved all
+                    technical tasks within project deadlines.</p>
+                    <p> Supervised, created and updated solutions that integrated with Oracle ERP System for a variety of modules, and handled
+                    integration with other systems.</p>
+            
+                </div>
```
LucyMac, line 179: I think all these text blocks should be full width (`col-12`) on mobile

PR #46, opened 2020-12-29
London Class-7-Negin Dn-Responsive Grid Project-HTML/CSS Week 4

NeginDn:
Your Details
Your Name: Negin Dn
Your City: London
Your Slack Name: Negin Dn

Homework Details
Module: HTML/CSS Boostrap
Week: 4


index.html
```diff
@@ -5,16 +5,154 @@
...
+    <link href="https://fonts.googleapis.com/css2?family=Quicksand:wght@300&display=swap" rel="stylesheet">
+    <link href="https://fonts.googleapis.com/css2?family=Libre+Baskerville&display=swap" rel="stylesheet">
+
+
+    
     <title>Responsive grid project</title>
 </head>
-
 <body>
+
```
louisechow, line 28: It would be good for you to use semantic tags in your html. This helps to make your code more readable since you can easily understand the context of the content.

Some tags that I think you could use are; `header`, `main`,`section`, `footer`. You can see more tags here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element

index.html
```diff
@@ -5,16 +5,154 @@
...
+            <div class=" col-12 col-md-4 head-txt">
+              <p>Maria Hope, Ph.D. - Psychologist</p>
+            </div>
+          </div>
+          </div>
+  
+  <!-- nav bar-->
+  <nav class="container-fluid">
+    <div class="row no-gutters menu-container">
+          <div class="offset-md-6 col-md-1"><a class="menu-item" href="./index.html"><p>Home</p></a></div>
```
louisechow, line 49: You can remove the `<p>` tags for your links, they aren't necessary here because you already created the `<a>` tag.

css/main.css
```diff
@@ -1,3 +1,213 @@
...
+}
+
+.article-list {
+  background-color: white;
+}
+.article-container p {
+  font-size: 25px;
+}
+
+.article-btn {
```
louisechow, line 189: This class is exactly the same as `app-btn`. When this happens, it means that you can just use one class in two places. 
What name would be good for this button class?

css/main.css
```diff
@@ -1,3 +1,213 @@
...
+  font-size: 25px;
+  font-family: Libre Baskeville;
+  text-align: center;
+  font-weight: lighter;
+  font-style: italic;
+  padding-top: 60px;
+  padding-bottom: 0;
+}
+
+.about-heading {
```
louisechow, line 122: Again, this is exactly the same as `intro-heading`. How can you make this code cleaner and reduce duplication?

index.html
NeginDn: Hi Louise,
Thank you for reviewing my code. I've changed it, It's not perfect yet and I need to improve my knowledge of different tags, but I change it as much as I could.

Thank you

index.html
NeginDn: Done

css/main.css
NeginDn: I just realise it's better to have shorter Css never thought of it before, I have sorted this out and next time coding I consider it. Thank you

css/main.css
NeginDn: Done

index.html
```diff
@@ -4,17 +4,136 @@
...
 </head>
-
 <body>
+  <header class="container header">
+    <div class="row ">
+      <button class=" offset-1 col-7 offset-md-9 col-md-2 standard-btn" type="button">Book an Appointment</button>
+      <div class=" offset-2 col-2 menu-icon">
+        <img src="./img/menu-hamburger.svg" alt="Hamburger menu">
+      </div>
+        <div class=" col-12 offset-md-1 col-md-4 head-txt">
```
louisechow, line 27: Don't forget to align sibling elements on the same time. Otherwise it makes the code difficult to read.

index.html
```diff
@@ -4,17 +4,136 @@
...
+      </div>
+      
+  <!-- nav bar-->
+  <nav class="row no-gutters menu-container">
+          <div class="offset-md-6 col-md-1"><a class="menu-item" href="./index.html">Home</a></div>
+          <div class="col-md-1 menu-item"><a  class="menu-item"href="./about.html">About</a></div>
+          <div class="col-md-1 menu-item"><a  class="menu-item"href="./service.html">Service</a></div>
+          <div class="col-md-1 menu-item"><a  class="menu-item"href="#Blog">Blog</a></div>
+          <div class="col-md-1 menu-item"><a  class="menu-item"href="./contact.html">Contact</a></div>
+      </div>
```
louisechow, line 39: These two closing divs are orphans - they don't have a matching partner :(
I think you can remove them

index.html
```diff
@@ -4,17 +4,136 @@
...
+        <img src="./img/menu-hamburger.svg" alt="Hamburger menu">
+      </div>
+        <div class=" col-12 offset-md-1 col-md-4 head-txt">
+          <p>Maria Hope, Ph.D. - Psychologist</p>
+        </div>
+      </div>
+      
+  <!-- nav bar-->
+  <nav class="row no-gutters menu-container">
+          <div class="offset-md-6 col-md-1"><a class="menu-item" href="./index.html">Home</a></div>
```
louisechow, line 34: I would suggest using using list tags for your navigation menu rather than `<div>`. It's more appropriate semantically.

E.g.
```
<nav>
  <ul>
    <li class=""><a href="#">Home</a></li>
    <li class=""><a href="#">About</a></li>
  </ul>
</nav>
```

index.html
```diff
@@ -4,17 +4,136 @@
...
+          <div class="col-md-1 menu-item"><a  class="menu-item"href="#Blog">Blog</a></div>
+          <div class="col-md-1 menu-item"><a  class="menu-item"href="./contact.html">Contact</a></div>
+      </div>
+    </div>
+  </nav>
+</header>
+
+   <!--inro-->
+<main>
+      <div class="container intro-container">
```
louisechow, line 46: I would suggest using he `<section>` tag for each section in your page, again it adds semantic organisation to your code.

css/main.css
```diff
@@ -1,3 +1,209 @@
...
+.standard-btn {
+  background-color: #cff4ec;
+  border: 0px;
+  padding-bottom: 10px;
+  padding-top: 10px;
+  margin-top: 20px;
+  margin-bottom: 20px;
+}
+
+.heading {
```
louisechow, line 39: You could simply your code and remove the need to add the `heading` class to all your h1 tags. Define the styles on the h1 tag instead.

E.g.
h1 {
  font-family: Libre Baskeville;
  font-size: 55px;
}

louisechow:
Nice work! As I mentioned to you before I think you've structured your code well (elements are well nested) and you've used bootstrap well.

If you have more time to spend you can practice more and implement the article section. It has a carousel in the mobile view which will give you experience using the bootstrap carousel function.

And don't forget to indent and format your code well. Otherwise you increase your chance of accidentally orphaning tags (not having corresponding close tags for every open and vice versa).

PR #47, opened 2020-12-29
London Class-7-Omar Sharif-Responsive Grid Project-HTML/CSS Week 4

omarsharif-gr:


css/contact-form.css
```diff
@@ -0,0 +1,148 @@
...
+    border: 1px solid red;
+    border-radius: 4px;
+    box-sizing: border-box;
+    margin-top: 6px;
+    margin-bottom: 16px;
+    resize: vertical;
+}
+
+input[type=text]:focus,input[type=email]:focus, textarea:focus {
+	border: 1px solid green;
```
LucyMac, line 118: 2 things here: 
1) the design has a solid black **bottom** border only, which turns to blue when focused - can you try and recreate this?
2) inputs have another style by default, called 'outline'. This is what is giving the fields a brown border. You can set `outline` to `0` for this to disappear.

css/main.css
```diff
@@ -1,3 +1,209 @@
...
+  .main{
+    display: flex;
+    flex-direction: row;
+    justify-content: center;
+  }
+  .maintextposition{
+    padding-top: 0px;
+    padding-bottom: 70px;
+    box-shadow:20px 20px 50px grey;
+    background-color: white;
```
LucyMac, line 173: This Hello section is going all the way to the right edge of the screen because it is not contained inside a `col-...` class
As mentioned in other comments, move the container up to the parent, to wrap the _whole_ block (profile image and text). And also add a `row` and `col-12` at the top level before you set the white background. See my suggested code for your resume page for inspiration.

LucyMac (commented):
A good attempt and coded in a mobile-first way which is great 👍
I have made suggestions to improve / clean up your layout a little, and you could put some more work in following the given design more precisely.

PR #48, opened 2020-12-29
London Class-7-Oksana Kostenko-Responsive Grid Project-HTML/CSS Week 4

oksygenn:


index.html
```diff
@@ -1,20 +1,247 @@
...
+                class="social-img"
+                src="https://static.wixstatic.com/media/01c3aff52f2a4dffa526d7a9843d46ea.png/v1/fill/w_40,h_40,al_c,usm_0.66_1.00_0.01/01c3aff52f2a4dffa526d7a9843d46ea.png"
+                alt="instagram icon"
+              />
+            </a>
+          </div>
+        </div>
 
-<body>
+        <div class="d-flex flex-column mx-md-5 py-5 w-100">
```
LucyMac, line 157: Instead of using `w-100`, use `col-12` here. Make sure this div's parent has the `row` class and apply `col-...` classes to this section and the beige photo block (hero-container) above too.

contact.html
```diff
@@ -0,0 +1,180 @@
...
+        <span class="bg-primary rect d-inline-block me-2"></span>Let's talk
+      </h1>
+
+      <form action="" class="container form-container">
+        <div class="row bg-white p-3 p-md-5">
+          <div class="col-12 col-md-6 mb-4">
+            <label for="first-name"
+              >First Name <span class="text-primary">*</span></label
+            >
+            <input type="text" class="form-control" id="first-name" required />
```
LucyMac, line 88: Your form looks great but again, write your own CSS to follow the design exactly. When the input is active, it should have a bottom blue line, not a black box. 
You'll need the pseudo-element `:focus` to style this.

LucyMac (commented):
Nice work Oksana, your site looks really good and the responsive behaviour is great 👏 
Try and rely on Bootstrap **less** for styling and **more** for layout, using the grid system we learned in class. I've noticed you tend to make your own layouts using `w-100` etc. Stick to using `container` (or `container-fluid`), `row` and `col-...` for layout, and write the rest as your own custom CSS. 

I've linked an [example solution](https://codesandbox.io/s/resume-layout-solution-j1tw5?file=/index.html) for the homepage layout which I hope will help 😊

oksygenn:
> Nice work Oksana, your site looks really good and the responsive behaviour is great 👏
> Try and rely on Bootstrap **less** for styling and **more** for layout, using the grid system we learned in class. I've noticed you tend to make your own layouts using `w-100` etc. Stick to using `container` (or `container-fluid`), `row` and `col-...` for layout, and write the rest as your own custom CSS.
> 
> I've linked an [example solution](https://codesandbox.io/s/resume-layout-solution-j1tw5?file=/index.html) for the homepage layout which I hope will help 😊

Thank you very much for your feedback, Lucy! 😊
I misunderstood the goal here and thought that the more bootstrap classes i use the better :D that's why I styled lots of things with them. 
I already changed some things that you suggested, and will try to change the rest later. 😊

PR #50, opened 2020-12-29
London class 7/abdulaziz alhasan - responsive-grid-project-HTML/CSS Week 4

AbdulazizAlhasan:
Still require some work, submited due to the deadline.
Resume and Projects pages are built using bootstrap grid, while About and contact are build using other bootstrap elements


css/main.css
```diff
@@ -1,3 +1,89 @@
...
 
 /* Dont' forget to link this file to your HTML in the <head> */
+
+:root {
+    --beige: #E6DACD;
+    --lbeige: #F4ECE6;
+    --blue: #0150FD;
+    --white: #FFF;
+    --black: #000;
+}
```
LucyMac, line 11: Great work using CSS variables!

LucyMac: 💡 A little trick to avoid repeating the same CSS in all your CSS files: you could create a `common.css` file which you import in all your HTML pages, on top of each page's individual styles. This way you put all the common styles (such as colours, fonts etc) in one place and if you need to change them, the changes will be instantly rolled out across all pages.

Projects.html
```diff
@@ -0,0 +1,122 @@
...
+            </li>
+          </ul>
+        </div>
+      </div>
+    </nav>
+    <div id="navholder"></div>
+    <main>
+      <div>
+        <h1 class="d-flex justify-content-center mt-4 py-3 ">projects</h1>
+        <div class="row mb-5">
```
LucyMac, line 48: To avoid the space between the text and the images here, you can apply Bootstrap's `g-0` class to this row. This stands for "gutter zero" which removes the spaces between the columns.
(this was previously known as `no-gutters` in Bootstrap 4 but has been changed to `g-0` in Bootstrap 5)

Contact.html
```diff
@@ -0,0 +1,133 @@
...
+        <h1 class="d-flex justify-content-center my-5 py-3">Let's talk</h1>
+      </div>
+      <div class="d-flex justify-content-center">
+        <form class="card shadow border-0 rounded-0 mb-5 p-5 pb-1 " style="width: 680px; height: 700px;">
+          <!-- 2 column grid layout with text inputs for the first and last names -->
+        <div class="row mb-4">
+          <div class="col">
+            <div class="form-outline">
+              <label class="form-label" for="form3Example1">First name</label>
+              <input type="text" id="form3Example1" class="form-control" />
```
LucyMac, line 57: Don't rely (only) on Bootstrap to style these fields. The design has a bottom black border only, which turns blue on focus.

css/main.css
```diff
@@ -1,3 +1,89 @@
...
+#background {
+    position: fixed;
+    top: 0px;
+    left: 0px;
+    width: 42%;
+    height: 900px;
+    background-color: var(--beige);
+    z-index: 1;
+}
```
LucyMac, line 47: Don't use IDs in your CSS. You can use IDs for JavaScript, but when adding styles to HTML, target your elements using classes.

Resume.html
```diff
@@ -0,0 +1,166 @@
...
+      </div>
+    </nav>
+    <div id="navholder"></div>
+    <main>
+      <div>
+        <h1 class="d-flex justify-content-center mt-4 py-3 ">Resume</h1>
+        <div class="row mb-5">
+          
+            <h1 class="col-5 offset-1 col-md-4 offset-md-2 col-lg-3 offset-lg-3 col-xl-2 offset-xl-4 mb-5">Experience</h1>
+            <a href="#" class=" col-4 offset-1 col-md-3 col-lg-2  col-xl-1 btn btn-primary rounded-pill fw-bold mb-5 ">DOWNLOAD CV</a>
```
LucyMac, line 51: I think you could wrap these together in the same col class, and just set `display: flex` and `space-between` on it (with your own class) to align the text to the left and the button to the right.
It's best to avoid putting `col` classes directly on buttons and images - best to apply to a div wrapping the button/image.

LucyMac (commented):
Top marks for copying the design so precisely Abdulaziz, you show a great eye for detail 🌟 

Your layout on mobile needs some work though - and generally we want to be coding in a "mobile-first" way: this means you should write all your code to look good on mobile by default, and _then_ you add media queries (using **min**-width), as well as use Bootstrap's responsive grid classes, to adapt it for desktop. 
I can see here you have coded everything for desktop and overlooked mobile a bit... could you improve this? 

Here's a little mock up of the homepage to help with the background switch

PR #51, opened 2020-12-29
London class 7/Deniz Kelemence/responsive grid project-HTML/CSS Week 4  

ASLANDENIZ:


css/main.css
```diff
@@ -1,3 +1,139 @@
...
+}
+
+/* .li {
+display: flex;
+} */
+
+
+.hero-image {
+    background-image: url(https://images.unsplash.com/photo-1514533450685-4493e01d1fdc?ixid=MXwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHw%3D&ixlib=rb-1.2.1&auto=format&fit=crop&w=1934&q=80);
+
```
LucyMac, line 23: To follow the design here, you will need to overlay this image (in smaller size) with the large golden shape in the background, and the beige small shape on top. Fo this, you'll need to use `position: absolute` on some elements (not all of them, but probably the small beige shape at least) and also `z-index` to define which one is on top / underneath.

css/main.css
```diff
@@ -1,3 +1,139 @@
...
+    /* padding-top:5px; */
+    padding-bottom:10px;
+}
+
+.line-up-section {
+    background-image: url(../img/festival_lilly-flower.svg), url(../img/festival_blob-shape.svg);
+    background-repeat: no-repeat, no-repeat;
+    background-position: left bottom, center;
+    background-size: 90px, contain;
+    background-color: rgb(97, 153, 32);
```
LucyMac, line 59: A little creative licence is ok as this is just an exercise, but bear in mind that the task (and the job of a front-end developer!) is to re-create designs exactly, included colours...

index.html
```diff
@@ -1,20 +1,207 @@
...
+                    <h1>DEW FESTIVAL MAY 28</h1>
+                    <h2>May 28, 2023, 9:00 AM</h2>
+
+                    <button class="ticket-button" type="button" aria-placeholder="Buy Tickets">Buy
+                        Tickets</button>
+                </div>
+        </section>
+        <section class="container-fluid line-up-section">
+            <div class="row">
+                <div class="col-10 offset-1 col-md-8 offset-md-2">
```
LucyMac, line 74: Here, in order to separate the background shapes and be able to control them (to rotate them and size them appropriately), I suggest moving the background-images to the column level, and having 2 separate column blocks: one for the backgrounds, and one for the text. This way, you can rotate, size, style as you wish, without affecting the others:
```
    <section class="container-fluid line-up-section">
      <div class="row">
        <div class="background-shapes col-10 offset-1 col-md-8 offset-md-2"></div>
        <div class="content col-10 offset-1 col-md-8 offset-md-2">
          <h3>Line up</h3>
          <p>ANDRE'S BEST / BLAKE ADAMSON / CAMMILLE / DJ CANOPY / DELORA PLANK / BELL HARRIS / NADINE BELMONT etc. </p>
        </div>
      </div>
    </section>
```
The `background-shapes` class can deal with the background-shapes, and also apply `transform: rotate(135g)` to match the design.
You'll notice if you follow this, the shape will be one block, and the text, a separate block below. To remedy this, apply `position: absolute` to the `background-shapes` class, as well as a `z-index` to both blocks to define which one sits on top.

LucyMac: One more hint: `background-shapes` will need a `height` set for it to appear, as it will not have content of its own to grow the block.

css/main.css
```diff
@@ -1,3 +1,139 @@
...
+    background-size: 60px, contain;
+    padding-top: 100px;
+    padding-bottom: 100px;
+    text-align: center;
+    
+}
+
+.about {
+    background: url(../img/festival_blob-shape.svg) no-repeat, url(../img/festival_lilly-flower.svg) center bottom no-repeat;
+    background-size: 100px, 100px;
```
LucyMac, line 91: Again here, best to separate out your background images on to a separate `div` and manipulate them separately using `position: absolute`, `z-index` and `transform: rotate(90deg)`

LucyMac: A trick if you want your shapes to be different colours: duplicate the image SVG in your image folder, and change this property inside the SVG: `fill="#f9a889"`
This is what dictates the colour of the shape. You should name your new SVG files accordingly (e.g. `blob-shape-blue.svg`, `blob-shape-gold.svg` etc)

index.html
```diff
@@ -1,20 +1,207 @@
...
+            </div>
+
+
+        </section>
+
+        <section class="container-fluid about">
+            <div class="row">
+                <div class="col-10 offset-1">
+                    <h2>ABOUT</h2>
+                    <p>I'M A PARAGRAPH. CLICK HERE TO ADD YOUR OWN TEXT AND EDIT ME. IT'S EASY. JUST CLICK "EDIT TEXT"
```
LucyMac, line 113: I notice you haven't set your own font-sizes to the headings. Notice in the design, the 'ABOUT' heading is actually smaller than the paragraph text.

LucyMac (commented):
Nice start Deniz, your use of Bootstrap is spot on 👍 
The page seems unfinished - can you finish the hero section so it matches the design? It would also be great if you could try and pin the links 'tickets', line-up' etc to the screen (hint: position: fixed`) to re-create the desktop design. 
Also make font-sizes and spaces larger as required on desktop.

Keep going! 💪

css/main.css
ASLANDENIZ: Hello Lucy, 

Thank you so much! I have made all the changes that you have requested. It was a great project.

css/main.css
```diff
@@ -1,3 +1,878 @@
...
+    background-size: cover;
+}
+
+.flower-shape {
+     background-image: url(../img/festival_lilly-flower.svg);
+     background-repeat: no-repeat;
+     background-size: 63px;
+     height: 66px;
+     width: 50px;
+     left:228px;
```
LucyMac, line 196: Since you want the flower to be pinned to the right of the brown shape, you can set the `right` property instead of the `left` here. It should give you more control because at the moment the flower is not moving in line with the brown shape when I resize the screen.

css/main.css
```diff
@@ -1,3 +1,878 @@
...
+     width: 97px;
+     top: 103px;
+     left: 15px;
+     position:absolute;
+     transform:rotate(108deg);
+}
+
+.ticket-section {
+    /* background-color: rgb(160, 160, 197); */
+    background-image: url(../img/festival_blob-shape-gold.svg);
```
LucyMac, line 217: I recommend **not** setting the background image on the `container-fluid` element - I would have the `container-fluid` class on the `section`, and then inside have a `div` with the class of `ticket-section`.

The reason for this is that you want to pin your flower and smaller shape image to this large brown blob, but because this blob is the full width of the screen, you're actually positioning your other images to the edges of the screen rather than just the edges of the large shape.

If you reduce the large blob by moving it inside the container, it might help you to position the other 2 images.

I realise this may sound confusing - happy to chat about this if it's unclear.

css/main.css
```diff
@@ -1,3 +1,878 @@
...
+    z-index:1;
+}
+
+.flower-three {
+    background: url(../img/festival_lilly-flower.svg);
+    background-size: 66px;
+    background-repeat: no-repeat;
+    height: 100px;
+    width: 100px;
+    left: 205px;
```
LucyMac, line 282: Same comment here - pin the flower using `right` rather than `left`

index.html
```diff
@@ -1,20 +1,970 @@
...
+
+            </div>
+
+
+        </section>
+        <section class="container-fluid first-section" id="first">
+            <div class="row justify-content-center">
+                <div class="relative">
+
+                    <svg class="col-12 col-md-8 offset-md-2 big-img-ab" data-bbox="18.416 22.603 162.965 154.895"
```
LucyMac, line 104: If you inspect this image in mobile mode, you'll see that because of the rotation, it's become wider than the page and is pushing the sides out (when incpecting you can see the element highlighted as a blue box on the page - observe how why that blue box is!). This is why there is a horizontal scroll on your mobile page.
To fix this, set `overflow: hidden;` on the section. It will essentially chop off anything that is overlapping outside of the desired size.

LucyMac (commented):
Great improvements Deniz, good work!
I've made a few comments below 📝

PR #54, opened 2020-12-29
London Class 7-Benyam Worku-Responsive Grid Project-HTML/CSS-Week 4

BenyamWorku:


index.html
```diff
@@ -1,20 +1,197 @@
...
+    <script src="https://code.jquery.com/jquery-3.5.1.slim.min.js"
+        integrity="sha384-DfXdz2htPH0lsSSs5nCTpuj/zy4C+OGpamoFVy38MVBnE+IbbVYUew+OrCXaRkfj"
+        crossorigin="anonymous"></script>
+    <script src="https://cdn.jsdelivr.net/npm/bootstrap@4.5.3/dist/js/bootstrap.bundle.min.js"
+        integrity="sha384-ho+j7jyWK8fNQe+A12Hb8AhRq26LrZ/JpcUGGOn+Y7RsweNrtN/tE3MoK7ZeZDyx"
+        crossorigin="anonymous"></script>
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
+
+    <!--  -->
+    <!-- <script src="https://code.jquery.com/jquery-3.5.1.min.js" -->
```
amellie, line 22: you can remove these codes if they are not used

index.html
```diff
@@ -1,20 +1,197 @@
...
+        <!-- </script> -->
     <!-- Add a link to your custom CSS here -->
+    <link rel="stylesheet" type="text/css" href="css/main.css" title="Default Styles" media="screen">
+    <script src="main.js"></script>
 
-    <title>Responsive grid project</title>
 </head>
 
 <body>
+        <div id="body-wrapper">
```
amellie, line 37: I would recommend to use the ``body-wrapper`` in the ``<body>`` tag instead. Always use a class instead of an ID as this is the best practice

index.html
```diff
@@ -1,20 +1,197 @@
...
+            <!-- header -->
+            
+            <!-- <header class="index-header-class"> -->
```
amellie, line 40: It is always good to use a ``<header>`` tag. I am not sure why this is removed?

index.html
```diff
@@ -1,20 +1,197 @@
...
+                        <p>I’m a great place for you to
+                            tell a story and let your
+                            users know a little more
+                            about you.</p>
+                    </div>
+                </div>
+            </div>
+        <!-- Footer -->
+            <div id="footer-wrapper">
+                <footer>
```
amellie, line 143: you can use ``footer-wrapper`` in ``<footer>`` tag, so the div is not needed here

index.html
```diff
@@ -1,20 +1,197 @@
...
+                    <!-- Contacts container -->
+                    <div class="contact container">
```
amellie, line 145: you use 2 different containers here for Call/Write and Social Media. However, this makes the components not aligned into one row, as per the original design (all components need to be in one row). So what you actually need is one container, containing one row. This row will contain Copyright, Call, Write, and Social Media icons. The width for Copyright is wider than the rest, and this can be easily changed as per the doc https://getbootstrap.com/docs/4.0/layout/grid/#variable-width-content

css/main.css
```diff
@@ -1,3 +1,146 @@
...
+  background-color: white;
+}
+.social-media a {
+  text-decoration: none;
+  padding: 20px;
+}
+.social-media .fa {
+  color: black;
+  text-align: center;
+  font-size: 15px; /*this controls the font size*/
```
amellie, line 98: Sizing of icons can use Font Awesome's available classes. See https://fontawesome.com/how-to-use/on-the-web/styling/sizing-icons

css/main.css
```diff
@@ -1,3 +1,146 @@
...
+}
+body {
+  background-color: #e6dacd;
+}
+
+/* Menu Bar */
+.navbar-light {
+  padding-bottom: 100px;
+}
+nav > div > div > a {
```
amellie, line 23: I would recommend using a class here. If your design gets more complicated, you have a risk of having this style applied to different ``<a>`` tags by accident

index.html
```diff
@@ -1,20 +1,197 @@
...
+            <!-- </header>  -->
+            <!-- profile card :photo,buttons,social-media  -->
+            <div class=" profile-card container ">
+                <div class="photo-wrapper row">
+                    <div class="offset-3 col-6">
+                        <img src="img/Benny_photo.png" alt="Ben's photo">
+                    </div>
+                </div>
+                <div class="name-wrapper row">
+                    <div class="offset-4 col-5">
```
amellie, line 82: Currently, your name is not exactly centered. Using ``col`` class here allows to have your name centered

index.html
```diff
@@ -1,20 +1,197 @@
...
+                <div class=" row">
+                    <div class="separator-line offset-5 col-2">
+                        <!-- blue line -->
+                    </div>
+                </div>
+                <div class="buttons row">
+                    <div class="resume-btn   col-6">
+                        <!-- <button>RESUME</button> -->
+                        <a href="resume.html">
+                            <button type="button" class="btn btn-primary">
```
amellie, line 95: you can style your own button instead of using the default Bootstrap button here

css/main.css
```diff
@@ -1,3 +1,146 @@
...
+form {
+  margin-bottom: 15px;
+}
+
+.form-heading {
+  margin: 50px 25px auto;
+  text-align: center;
+}
+.contact-form input {
+  background: transparent;
```
amellie, line 117: I don't think this makes any difference

css/main.css
```diff
@@ -1,3 +1,146 @@
...
+  border-bottom: 1px solid black;
```
amellie, line 118: This is correct, however, the input box still has a grey top/right/left border which needs to be replaced in ``form-control`` class

contact.html
```diff
@@ -0,0 +1,170 @@
...
+                        <label for="email ">Email</label>
+                        <input type="email" name="email" id="email" class="form-control">
+                    </div>
+                    <div class="form-group">
+                        <label for="subject">Subject</label>
+                        <input type="text" name="subject" id="subject" class="form-control">
+                    </div>
+                    <div class="form-group">
+                        <label for="message">Message</label>
+                        <input type="text" name="message" id="message" class="form-control">
```
amellie, line 108: Message text box is wider and bigger than the rest of the text boxes. So you should be using ``textarea`` instead

contact.html
```diff
@@ -0,0 +1,170 @@
...
+        </div>
+        
+    </div>
+    <!-- form with multiple entries -->
+    <div class="form-container container">
+        <div class="row">
+            <div class="col-12">
+                <form  action="" class="contact-form" >
+                    <div class="form-group">
+                        <label for="first name">First Name</label>
```
amellie, line 91: First Name and Last Name should be in one row :)

css/main.css
```diff
@@ -1,3 +1,146 @@
...
+}
+
+.contact-form input:focus {
+  outline: 1px solid royalblue;
+  border: none;
+}
+
+.contact-form {
+  background: white;
```
amellie, line 127: how do you make sure that there is a space inside the white box? (between the content of the form and the edge of the white box)

index.html
```diff
@@ -1,20 +1,197 @@
...
+                        </div>
+                    </div>
+                </nav>
+                
+            <!-- </header>  -->
+            <!-- profile card :photo,buttons,social-media  -->
+            <div class=" profile-card container ">
+                <div class="photo-wrapper row">
+                    <div class="offset-3 col-6">
+                        <img src="img/Benny_photo.png" alt="Ben's photo">
```
amellie, line 78: you can resize this photo using CSS

amellie (changes requested):
Well done on your work, Ben. You use a fair amount of Bootstrap and your own CSS as well. Just be mindful when to use the container, row, and col, to ensure that your components are placed correctly. Nice responsive homepage as well! :)

BenyamWorku:
> Well done on your work, Ben. You use a fair amount of Bootstrap and your own CSS as well. Just be mindful when to use the container, row, and col, to ensure that your components are placed correctly. Nice responsive homepage as well! :)

Thanks Amalia!

PR #55, opened 2020-12-30
London Class 7 - Soraiya Muhammad - Responsive Grid Project - HTML/CSS - Week 4

soraiya:


index.html
```diff
@@ -7,14 +7,173 @@
...
+                great place for you to tell a story and let your users know a little more about you.</p>
+            </div>
+        </div>
+        <div class="row mb-3 arrow-row">
+            <div class="col-1 offset-6 arrow-column">
+                <button class="arrow">→</button>
+            </div>
+        </div>
+        <div class="row justify-content-center bio-img-row">
+            <div class="col-5 bio-img-col">
```
LucyMac, line 83: Trick to get the overlapping image effect:
Add `position: absolute` to this line (line 83)
And then add this to the section:
```
.bio-section {
   margin-bottom: 300px;
   padding-bottom: 400px;
}
    ```
(You might also need to add `z-index: 1;` to your `nav` as the woman's image now sits on top of it)

File: css/main.css [hunk 2]
```diff
@@ -1,3 +1,218 @@
...
+.row.blog-images {
+  background-color: white;
+  width: 90%;
+  margin: 30px 0px 30px 45px;
+
+}
+
+.blog-images img {
+  width: 100%;
+  height: 200px;
```
LucyMac, line 200: Don't set a width AND a height on your `img` as this will distort them. `width: 100%` should be enough, and let the height set itself naturally.
(You may need to get images that are the same aspect ratio as each other so they all line up)

LucyMac (commented):
Nice start Soraiya, you've shown a good understanding of how Bootstrap works. You've done this in a desktop-first way however (meaning your code works for desktop by default), and more importantly there are no media queries / it's not responsive 😢 
Can you make changes so your code works on mobile by default, and then use Bootstrap's responsive grid classes as well as a `min-width` media query to make changes for desktop? 
I'd also like to see more than one page - especially the 'contact' page would be good practice to work with forms.

PR #56, opened 2020-12-30
London Class 7 - Mustafa Ak - Responsive Grid Project- HTML/CSS - Week 4 

akbasmustafa:
Your Details
Your Name: Mustafa Ak
Your City: London
Your Slack Name: Mustafa Ak
Homework Details
Module: HTML/CSS-Responsive Grid Project
Week: 4
Link: https://cyf-akbasmustafa-responsivegridproject.netlify.app/


css/main.css
```diff
@@ -1,3 +1,45 @@
...
+    background-color: #faefe9;
+    width: 46%;
+    height: 100%;
+  }
+
+  .main-background div:nth-child(2) {
+    background-color: white;
+    width: 54%;
+    height: 100%;
+  }
```
LucyMac, line 44: Good work making the dark beige background work on mobile and on desktop 👍 Watch out for the height of that background on mobile though - the height is not adapting to the content. It's a little tricky to get this to work, have a look at [my sample solution here](https://codesandbox.io/s/resume-layout-solution-j1tw5?file=/index.html).

LucyMac: Also, using combined selectors like this `div:nth-child(2)` is fine here, but can get complicated when you have more complex markup. Don't hesitate to add your own custom class here for this (like I did with 'pinned-background')

css/mainresume.css
```diff
@@ -0,0 +1,46 @@
+.main-content {
+  position: absolute;
+  top: 8rem;
+  left: 10%;
+  width: 90%;
+}
```
LucyMac, line 6: You are using this in conjunction with the class `row` - this defeats the purpose of using the grid system because you are setting your own widths and margins. This is Bootstrap's job!
This is why your content is not centered, because your sizing/spacing is clashing with Bootstrap's grid system. 
When applying grid classes to elements (`container`, `row`, `col-...`), you can add your own custom classes of course, but only for colours, text, **vertical** spacing etc - not for  horizontal spacing such as widths or left/right padding or margin.

projects.html
```diff
@@ -0,0 +1,252 @@
...
+          <p>
+            I'm a paragraph. Click here to add your own text and edit me. It’s
+            easy. Just click “Edit Text” or double click me to add your own
+            content and make changes to the font. I’m a great place for you to
+            tell a story and let your users know a little more about you.
+          </p>
+        </div>
+        <div class="col-12 col-md-7 mb-5 content-card">
+          <div class="row justify-content-between align-items-center">
+            <div class="col-12 col-md-6">
```
LucyMac, line 130: To ensure your white cards don't touch the edges of the screen, try moving `content-card` down, like this: 
```
        <div class="col-12 col-md-7">
          <div class="row">
            <div class="col-12 col-md-6 content-card">
```

projects.html
```diff
@@ -0,0 +1,252 @@
...
+              <p class="h3" style="color: blue; font-weight: bold;">
+                Project name 01
+              </p>
```
LucyMac, line 133: 2 problems here:
1) Don't use inline styling, keep your styles in your CSS file. This is both for maintainability and because inline styles 'win' the [specificity battle](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity), which can cause you trouble in the future if you try to override these. 
2) Why choose a `<p>` tag if it could be an `h3` tag? Remember to pick your HTML tags semantically.

resume.html
```diff
@@ -0,0 +1,345 @@
...
+  </head>
+
+  <body class="body">
+    <header class="container-fluid">
+      <div class="row">
+        <nav
+          class="navbar navbar-expand-md fixed-top"
+          style="background-color: white; height: 5rem;"
+        >
+        <a class="col-6 col-md-5 navbar-brand" href="#">
```
LucyMac, line 48: Since `row` sets `display: flex` it needs the `col-...` elements to be its direct children in order to work properly. This means you should always follow `row` with `col-...` (here your nav is in between the two, breaking the direct relationship)

contacts.html
```diff
@@ -0,0 +1,215 @@
...
+          </h5>
+        </div>
+        <div class="col-12 col-md-7 content-card">
+          <div class="row justify-content-between align-items-center">
+            <div class="col-12 col-md-6 p-4">
+              <label for="nameinput"
+                >First Name
+                <span style="color: blue; font-size: 1.3rem;">*</i>
+              </label>
+              <input type="text" id="nameinput" class="form-control" />
```
LucyMac, line 130: Don't rely on Bootstrap for anything other than layout. Bootstrap can help with giving you a starting point for elements like form inputs or buttons, but you must still add your own CSS to re-create the design provided to you.
The form fields in the design look different, they have a bottom black border which turns bright blue on hover. Add your own custom class to these fields to re-create the design.

contacts.html
```diff
@@ -0,0 +1,215 @@
...
+          </div>
+        </div>
+      </div>
+    </main>
+
+    <footer class="container-fluid">
+      <div class="row m-3 align-items-center">
+        <div class="col-12 col-md-8 order-md-2 mb-3">
+          <div class="row">
+            <div class="col-6 col-md-4 p-1 text-center">
```
LucyMac, line 170: Don't set different left/right padding or margin on elements that have Bootstrap grid classes on them (i.e. `container`, `row` or `col-..`). These Bootstrap classes have their own specific spacing which is what makes the grid work in a consistent manner. If you override this with your own padding/margin, it clashes with the grid system and can cause misalignment on your page. 

You can of course always add your own top and bottom padding or margin, as the grid only deals with horizontal spacing. Here, `p-1` adds padding on all sides which goes against the grid.

css/mainresume.css
```diff
@@ -0,0 +1,46 @@
...
+    /* left: 25%; */
+    /* max-width: 50%; */
+    min-width: 700px;
+  }
+  .main-background {
+    height: 2200px;
+  }
+  .content-card {
+    min-width: 700px;
+  }
```
LucyMac, line 33: Using width when you are also using `col-...` classes on the same element is an issue. Choose one or the other 🙂

LucyMac (commented):
Nice work Mustafa, your site looks really good and you've overcome some tricky layout challenges in CSS 👍 
Just be careful to follow the rules of Bootstrap: don't set custom widths or left/right padding or margins on elements that use the grid system (either `container`, `row` or `col-...` elements)

PR #57, opened 2021-01-07
London Class 7 - Ionut Tripa - Responsive Grid Project - HTML/CSS Week 4

iTrip23:


index.html
```diff
@@ -7,7 +7,7 @@
     <!-- Here we import the Boostrap CSS stylesheet -->
     <link href="https://cdn.jsdelivr.net/npm/bootstrap@5.0.0-beta1/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-giJF6kkoqNQ00vy+HMDP7azOuL0xtbfIcaT9wjKHr8RbDVddVHyTfAAsrekwKmP1" crossorigin="anonymous">
     <!-- Add a link to your custom CSS here -->
-
+    <link rel="stylesheet" href="css/main.css">
     <title>Responsive grid project</title>
 </head>
```
LucyMac, line 13: Waiting for some homepage content here ⏳

LucyMac (commented):
Johnny I know you can do better than this 😉

therapy/therapy-about.html
```diff
@@ -0,0 +1,160 @@
...
+                </ul>
+            </nav>
+            <div>
+                <h3 class="headings headings-therapy"><a href="therapy.html">Maria Hope, Ph.D. - Psychologist</a></h3>
+            </div>
+        </div>
+    </header>
+    <section class="container">
+        <div class="row">
+            <div class="about">
```
LucyMac, line 61: This is where you would use a col-... class. See comment below

therapy/therapy-about.html
```diff
@@ -0,0 +1,160 @@
...
+                <p class="header-p about-text about-p">
+                    I'm a paragraph. Click here to edit and add your own text. Is easy! Just click on "Edit Text" or double-click on me and
+                    you can add your own content and change fonts. Feel free to drag and drop anywhere on your page. I'm a great place for
+                    you to tell your story and let your customers know a little more about you. This is a great place to write a long text
+                    about your company and its services. You can use this space to go into details about your business. Talk about your team
+                    and about the services you provide. Tell your visitors about how you came up with the idea of starting your business and
+                    what makes it different from your competitors. Make your business stand out and show who you are.
+                </p>
+                <img class="col-12 about-maria" src="https://static.wixstatic.com/media/9c608a_2a78cf5937414d0e9a597504d68daafc~mv2_d_6046_4035_s_4_2.jpg/v1/fill/w_365,h_446,al_c,q_80,usm_0.66_1.00_0.01/9c608a_2a78cf5937414d0e9a597504d68daafc~mv2_d_6046_4035_s_4_2.webp" alt="maria-pic">
```
LucyMac, line 72: If you look at the CSS behind Bootstrap classes, you'll see that 'row' has 'display: flex', which it uses to control the layout of its flex children: the 'col-...' classes.
Therefore, for Bootstrap to work, any 'col-' class must be applied to a **direct** child of 'row'.
In this case, the direct children of the row on line 60 are the 'about' div on line 61 and the 'experience' div below - so they should have a 'col-...' class applied to them, not the img.

therapy/therapy-blog.html
```diff
@@ -0,0 +1,300 @@
...
+                </ul>
+            </nav>
+            <div>
+                <h3 class="headings headings-therapy"><a href="therapy.html">Maria Hope, Ph.D. - Psychologist</a></h3>
+            </div>
+        </div>
+    </header>
+    <section class="container">
+        <div class="row">
+            <div class="section-intro">
```
LucyMac, line 59: As per comments above, this element as well as 'blog-articles' below should have a column class applied to them as they are direct children of 'row'

therapy/therapy-about.html
LucyMac: For mobile, using 'col-12' is common as most things are full width. But what about for tablet or desktop? If the content is centered but less wide, you could use 'col-md-10' plus add an offset which will move it by 1 column to ensure it's centered: 'offset-md-1'

So in this example, you would apply all the following classes to the 'about' div: 'about col-12 col-md-10 offset-md-1'

LucyMac (commented):
Good start Johnny - see my comments about using the 3rd part of bootstrap's grid: columns 
See if you can add some col-... classes after your 'row's to set different widths to each block depending on the device

PR #58, opened 2021-01-08
London class 7-Aisha-khan-week 3

aishak05:


index.html
```diff
@@ -6,15 +6,164 @@
...
 
+
 <body>
+<div id="page">
+  
+    <header class="container">
+        <div class="row">
+            <nav class="navbar navbar-expand-sm fixed-top">
+
+                <div class="col-12 col-sm-4">
```
LucyMac, line 22: To follow correct use of Bootstrap, the `col-...` elements need to be direct children of `row`. Here, the `nav` is in between which is breaking that direct relationship. 
The reason for this is that `row` uses `display: flex`, which only works on its direct (i.e. immediate) children.

index.html
```diff
@@ -6,15 +6,164 @@
...
+                <h4>03.</h4>
+                <p><b>Couples Therapy</b></p>
+                <p>Lorem ipsum dolor sit amet consectetur adipisicing elit. Molestiae, voluptatibus quae, voluptatum veniam eveniet consequatur voluptate rerum dolorem suscipit ullam repudiandae. Labore, perspiciatis rerum minima praesentium recusandae molestiae eaque pariatur?</p>
+            </div>
+        </div>
+
+        <div class="row justify-content-center main-row4 item-4">
+            <h3 id="articles-text">Articles about Psychology</h3>
+                <div class="col-md-2 ">
+                    <div class="card" style="width: 18rem;">
```
LucyMac, line 113: Don't set styling inline like this - again an issue with specificity. Keep your styles in your CSS file.

Also in this case, you'd be better off setting your `card` class to `width: 100%` and letting its `col-...` parent (on line 112) determine how wide it gets: for mobile, either `col-12` or `col-10 offset-1` for example, and then for tablet you might want `col-md-4` as there are 3 of them and you want your columns to always add up to 12. 

As mentioned above, if you remove the fixed heights on this blog section and the services section above it, elements should flow a little better on mobile.

index.html
```diff
@@ -6,15 +6,164 @@
...
+                <div class="col-12 col-sm-4">
+                    <h1 class="navbar-brand">Maria Hope, Ph.D. - Psychologist</h1>
+                </div>
+                <div class="col-12 col-sm-8 justify-content-center d-flex flex-sm-column align-items-sm-end">
+                    <a class="col-8 col-sm-5 col-md-4 col-lg-2 appt-btn" href="#">Book an Appointment</a>
+                    <button class="navbar-toggler" type="button" data-toggle="collapse" data-target="#navbarCollapse">
+                        <span class="navbar-toggler-icon"></span>
+                    </button>
+                <div class="hamburger-icon">
+                    <img src="https://lh3.googleusercontent.com/Z4fssiZXxUUhPB_dHCABmQQ-HyCFAu0FGOk7viZuDB8a0pg3D3MuG0OWfX1u-xMtZQeJ9g=s85" alt="hamburger icon">
```
LucyMac, line 31: This image is not showing for me 🙁

css/main.css
```diff
@@ -1,3 +1,263 @@
...
+} */
+
+.nav{
+  margin: 0;
+}
+
+.navbar{
+  padding: 20px;
+  background-color: white;
+  width: 93%;
```
LucyMac, line 55: You probably want your navbar to take up the full width of the page?

index.html
```diff
@@ -6,15 +6,164 @@
...
+                            </li>
+                        </ul>
+                    </div>   
+                </div>
+            </nav>
+        </div>    
+    </header>
+
+
+    <div class="comtainer-fluid main">
```
LucyMac, line 58: typo here, should be `container-fluid`

LucyMac (commented):
Good start Aisha, however it looks like you've coded everything up for desktop first and not looked much at the behaviour on mobile. 
Clean up / remove your fixed heights as per my comments and re-adjust your Bootstrap column classes to account for the mobile layout so that your elements make better use of the space.
It would also be great to see another page of the site - maybe the contact page to get some practice making forms? 😃

PR #59, opened 2021-01-11
London Class 7 - Kara Howard - Responsive Grid Project - HTML/CSS Week 4

kmh256:
# Your Details

- Your Name: Kara Howard
- Your City: London
- Your Slack Name: Kara Howard

# Homework Details

- Responsive Grid Project HTML/CSS
- Week: 4


index.html
```diff
@@ -7,14 +7,202 @@
...
-
+    <div class = "container">
+        <div class = "row">
+            <div class = "col">
+                <!--Book an Appointment Button-->
+                <button class = "buttons" id = "booking">
+                    Book an Appointment
+                </button>
+            </div>
+    <div class = "container">
```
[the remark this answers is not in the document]
LucyMac: Try and clean all this up by aiming for each block inside `main` to be its own `row`.
You can decide whether to wrap all of these `row`s inside one single `container` at the top (option A below) or whether to wrap each separate section/row inside its own `container` (before the row - see option B).

**Option A** - one container wraps everything
```
  <main class="container">
    <section class="row">
      <div class="col-12">
        <!-- one block here -->
      </div>
    </section>
    <section class="row">
      <div class="col-12">
        <!-- another block here -->
      </div>
    </section>
  </main>
  ```

**Option B** - a container for each block
```
  <main>
    <section class="container">
      <div class="row">
        <div class="col-12">
          <!-- one block here -->
        </div>
      </div>
    </section>
    <section class="container">
      <div class="row">
        <div class="col-12">
          <!-- another block here -->
        </div>
      </div>
    </section>
  </main>
  ```

index.html
```diff
@@ -7,14 +7,202 @@
...
+                        <a class = "nav-link" href = "#">Contact</a>
+                     </li >
+                </ul>
+            </div>
+        </nav>
+    </div>
+    <div class="container text-center g-0">
+        <div class="row">
+            <div class="col-md-6 order-md-5 g-0">
+                <img class = "img-fluid" id="flower" src = "./img/green-flower.png" alt = "Green-Blue Flower">
```
LucyMac, line 60: You shouldn't need an `id` here

index.html
```diff
@@ -7,14 +7,202 @@
...
+            <div class = "col-md-4 g-0">
+                <section class = "individual">
+                    <h3>01.</h3>
+                    <h2>Individual Therapy</h2>
+                    <p>
+                        I'm a paragraph. Click here to add your own text and edit me. It's easy. Just click "Edit Text" or double click me to add your own content and make changes to the font.
+                    </p>
+                </section>
+            </div>
+            <div class = "col-md-4 g-0">
```
LucyMac, line 111: This seems to work fine here but FYI I believe you can apply the `g-0` class directly on the `row` so you don't have to repeat it on each `col-...` element

index.html
```diff
@@ -7,14 +7,202 @@
...
+                <section class = "group">
+                    <h3>02.</h3>
+                    <h2>Group Therapy</h2>
+                    <p>
+                        I'm a paragraph. Click here to add your own text and edit me. It's easy. Just click "Edit Text" or double click me to add your own content and make changes to the font.
+                    </p>
+                </section>
+            </div>
+            <div class = "col-md-4 g-0">
+                <section class = "couples">
```
LucyMac, line 121: I notice on desktop these 3 coloured blocks aren't all the same height. That's because they adapt to their content (i.e. less text needs less height). 
The 3 `col-md-4` blocks _are_ all the same height however because they are flex children and flex children always have the same height (their parent `row` has `display: flex`) so you can just set `height: 100%` to the coloured blocks (`individual `, `group` and `couples`) so they will grow to the height of their respective parents (which is `col-md-4`).
Hope that makes sense - happy to chat about this is it's not clear!

index.html
```diff
@@ -7,14 +7,202 @@
...
+                    <button class = "buttons">
+                        More Articles
+                    </button>
+                </div>
+            </section>
+        </div>
+    </div>
+    <!--Social Media-->
+    <div class = "container">
+        <footer>
```
LucyMac, line 182: Remember you don't have to apply Bootstrap classes to `div`s only - you don't need the div on line 181 here, you can just apply the `container` class to the `<footer>` directly

css/main.css
```diff
@@ -1,3 +1,96 @@
...
+    padding: 10px;
+}
+.about-maria {
+    padding-bottom: 150px;
+}
+.maria-photo {
+    margin-top: -120px;
+    padding-bottom: 75px;
+    z-index: 1;
+}
```
LucyMac, line 77: This works great, well done!

index.html
```diff
@@ -7,14 +7,202 @@
...
+            </button>
+            <div class = "collapse navbar-collapse" id = "navbarSupportedContent">
+                <ul class = "navbar-nav ml-auto">
+                    <li class = "nav-item">
+                        <a class = "nav-link" id = "home" href = "#">Home</a>
+                    </li >
+                    <li class = "nav-item">
+                        <a class = "nav-link" href = "#">About</a>
+                     </li >
+                     <li class = "nav-item">
```
LucyMac, line 44: We tend not to have spaces around the `=` sign in HTML attributes

css/main.css
```diff
@@ -1,3 +1,96 @@
...
+    margin: 30px 0 20px 10px;
+    font-family: 'Quicksand', sans-serif;
+}
+.buttons:hover {
+    background-color: #4D7369;
+    color: white;
+}
+#booking{
+    float: right
+}
```
LucyMac, line 17: Please don't use `float` rules in your CSS. Float is the old way of making layouts in CSS and it has now been replaced by Flexbox (which was introduced a few years ago). 
Set `display: flex` on the parent if you want to control where this element sits.

css/main.css
```diff
@@ -1,3 +1,96 @@
...
+    padding: 30px;
+}
+.individual {
+    background-color: #E2F6CD;
+    padding: 30px;
+}
+.group {
+    background-color: #F2F2F1;
+    padding: 30px;
+}
```
LucyMac, line 56: Remember to make CSS classes as re-usable as possible. This means you could have one class for the styles these blocks have in common, and separate classes for what differentiates them. 
Example:
```
.service-block { 
   padding: 30px;
}
.light-blue { 
   background-color: #CFF4EC;
} 
.light-grey { 
   background-color: #F2F2F1;
} 
.light-green { 
   background-color: #E2F6CD;
} 
```
and then you can set them in your HTML like this:
```
  <section class="service-block light-blue">
    <!-- blue content -->
  </section>

  <section class="service-block light-grey">
    <!-- grey content -->
  </section>

  <section class="service-block light-green">
    <!-- green content -->
  </section>
```

LucyMac (commented):
Great rendition of the design Kara, you have a good eye! 🎨 
A couple of issues with your HTML structure that need some tidying up - see comments below. 
Once you've fixed those, could you add another page or two to complete the assignment? The contact page especially would be good practice to get experience with form fields 😊 
Keep up the good work 👍

PR #60, opened 2021-01-16
london-class-7/Ramla-Jama/HTML- CSS/week 4

ramla199:


index.html
```diff
@@ -7,13 +7,109 @@
...
+             </div>
+          </nav>
+         </div>
+       </div>
+     </div>
+   </header>
+
+
+   <main class="container-fluid">
+       <div class="row sec1">
```
LucyMac, line 43: Can you use the more semantic `section` tag foe your top level blocks instead of a `div`?

LucyMac (commented):
Nice solution of the desktop design Ramla 👍 
Can you make it work for mobile?

Be careful with setting fixed heights to elements - it's usually best to let the content inside elements grow their container naturally to avoid unexpected behaviours at different screen sizes. Try removing the heights you've set to sections in your code and see what the impact is - sometimes, you might need them, but often you don't 🙂 

It would be great if you could add another page to your site - for example the contact page provides a good opportunity to practice styling forms 😉

PR #61, opened 2021-01-17
London class 7 - Somayeh Karimi- Responsive Grid Project-HTML/CSS Week4

DibaKarimi:


Buroni:
Great job Somayeh! Good use of Bootstrap grid and semantic HTML. You've also shown a good understanding of CSS `position`s by using `relative` and `fixed` when needed. Good work also applying the text and button hover/click effects.

One note for improvement is that you have a lot of duplicated CSS styles that exist in both `main.css` and `contact.css`. If you need these styles in both pages, why don't you create an extra CSS file for shared styles? Then you can include that shared file in both pages.

Overall very good work!

PR #62, opened 2021-01-22
London class 7   daniel ghanbari   responsive grid project   html/css   week 4

DanGhanbari:


Buroni (commented):
Good job @DanGhanbari! Nice use of media queries, bootstrap grid and flex to get a responsive layout that looks good on multiple display sizes. 

One minor style point is that the header is fixed in the Wix template, but it scrolls away in your website. See if you can change your header to be fixed too (should be a quick job).

I think you could have spent a bit more time styling your contact page -- the form looks too wide and you've forgotten to add the map image.

Overall great job. Not much to comment on the HTML or CSS side 👍

css/main.css
```diff
@@ -1,3 +1,49 @@
 /* Add your custom CSS here */
 
-/* Dont' forget to link this file to your HTML in the <head> */
```
Buroni, line 3: `italic` isn't a good class name; it's not only italic, but the font family and size are different.

css/main.css
```diff
@@ -1,3 +1,49 @@
...
+  font-size: 0.8rem;
+}
+.img-cover {
+  width: 100%;
+}
+
+@media (min-width: 768px) {
+  .order {
+    order: 2;
+  }
```
Buroni, line 39: You need to set `display: flex` if you want to use `align-items` and `justify-content`

contact.html
```diff
@@ -0,0 +1,127 @@
...
+            I'm a paragraph. Click here to add your own text and edit me. I’m a
+            great place for you to tell a story and let your users know a little
+            more about you.
+          </p>
+        </div>
+      </div>
+    </div>
+    <div class="row" style="background-color: #e2f7cd">
+      <div class="col">
+        <div class="container">
```
Buroni, line 53: You can write `py-4` instead of `pt-4 pb-4`.

contact.html
```diff
@@ -0,0 +1,127 @@
...
+                class="form-control"
+                id="exampleFormControlTextarea1"
+                rows="5"
+                placeholder="Message"
+              ></textarea>
+            </div>
+            <div class="d-grid gap-2">
+              <button class="btn btn-primary" type="button">Send</button>
+            </div>
+          </form>
```
Buroni, line 94: Try to write all your CSS styles inside your CSS file instead of using the HTML `style` attribute.

PR #64, opened 2021-12-07
London-class-8/craig-dsilva

craig-dsilva:


css/faq.css
```diff
@@ -0,0 +1,43 @@
+:root {
+  --black: #000;
+  --white: #fff;
+  --red-brown: #a05941;
+  --light-grey: #ededee;
+  --green-black: #133032;
+}
```
LucyMac, line 7: Nice - but do you need to re-declare these in each CSS file? How could you declare them once and share them across the project?

css/main.css
```diff
@@ -1,3 +1,345 @@
...
+  --light-grey: #ededee;
+  --green-black: #133032;
+}
+
+body {
+  background-color: var(--light-grey);
+  color: var(--red-brown);
+  margin: 0;
+  font-size: 20px;
+  font-family: "Fjalla One", sans-serif;
```
LucyMac, line 17: There is a bit of a mixup with the font families - compare you headings with the design, and then compare the body copy with the design. Do you see the font families have been inverted? Swap them around and you'll be much closer to the design 😉

css/main.css
```diff
@@ -1,3 +1,345 @@
...
+
+.search {
+  background-color: var(--light-grey);
+  border: 1px solid var(--red-brown);
+}
+
+.shopping-cart,
+.hamburger {
+  width: 50px;
+  height: 50px;
```
LucyMac, line 36: These seem a little big?

css/main.css
```diff
@@ -1,3 +1,345 @@
...
+  padding: 1rem 0;
+  margin: 1rem 0;
+  border: 1px solid var(--white);
+}
+
+.subscribe-now {
+  background-color: var(--white);
+  border: none;
+  padding: 1rem 0;
+}
```
LucyMac, line 241: If you imagine coding up this entire website, you are likely to need a similar text input, and a similar button somewhere else. Could you rename these classes to be more generic, and therefore more re-usable? For example `.text-input` and `button-primary`

css/main.css
```diff
@@ -1,3 +1,345 @@
...
+    grid-template-columns: 1fr 4fr auto;
+  }
+
+  .search {
+    grid-column: 1 / 3;
+  }
+
+  .shopping-cart {
+    grid-column: 2 / 3;
+  }
```
LucyMac, line 267: Because of how you have structured your HTML, `shopping-cart` isn't actually a grid item. It needs to be a **direct** child of the grid container in order for it to work. Move your class up to the `<a>` above it.

LucyMac: The same thing applies to `header-title`. That class needs to be moved up one level for its grid rules to work.

css/main.css
```diff
@@ -1,3 +1,345 @@
...
+    height: 70vh;
+  }
+
+  .about {
+    grid-template-columns: repeat(4, 1fr);
+  }
+
+  .about-text {
+    grid-column: 1 / 3;
+    margin: 0 10rem;
```
LucyMac, line 301: This margin is messing with the layout - remove it to let the text use up the full 2 columns

css/main.css
```diff
@@ -1,3 +1,345 @@
...
+  display: grid;
+}
+
+.about-text {
+  margin: 0 1.5rem;
+}
+
+.about-image {
+  width: 100%;
+}
```
LucyMac, line 144: These images are not covering the whole grid cell because they are not tall enough. To make them 'cover' the entire space, you have 2 choices:
1) Keep it as an `img` but add `object-fit: cover;` and `height: 100%;` (this will not work in IE)
2) Change the img to be a CSS `background-image` instead and use `background-size: cover`

css/main.css
```diff
@@ -1,3 +1,345 @@
...
+  color: var(--black);
+  background-color: var(--white);
+}
+
+.form-text {
+  text-align: center;
+  margin-bottom: 2rem;
+  color: var(--red-brown);
+}
+
```
LucyMac, line 184: The text inputs could do with a little padding to match the design. Also, first name and last name should be side by side on desktop.

LucyMac (commented):
Good work on the whole Craig. A few visual inconsistencies with the design (font families and spacing mostly) which can be easily fixed. 
Watch out when using CSS Grid - only direct children of the grid container will be considered as grid items, so any elements nested deeper in the structure will not work with rules like `grid-column` etc. See my comments below.

PR #65, opened 2021-12-08
London/8-Mahri-Atdayeva-Juice-Bar-final-HTML/CSS-project

mahri09:
completed html structure of website


index.html
```diff
@@ -1,21 +1,486 @@
...
+                ></path>
+              </g>
+            </svg>
+          </span>
+          <span>Log In</span>
+        </a>
+      </header>
+      <main>
+        <section class="home sections-container">
+          <h1><span>Fuel Your Body.</span><span>Balance Your Mind.</span></h1>
```
LucyMac: It's usually not best practice to force text to split like this. Let it break naturally. You could use padding to make sure it doesn't span too wide.

index.html
```diff
@@ -1,21 +1,486 @@
...
+        </section>
+        <section class="about-us sections-container">
+          <div class="aboutUs-image">
+            <img src="./images/fruit.jpg" alt="about-us-fruits-image" />
+          </div>
+          <article class="about-text">
+            <h1>
+              <span>Boosting</span> <span>Immunity the Way</span>
+              <span>Nature Intended</span>
+            </h1>
```
LucyMac: Again, just let the text break naturally so it can break as it needs to depending on the width of the screen.

index.html
```diff
@@ -1,21 +1,486 @@
...
+            </li>
+          </ul>
+        </div>
+        <form action="#">
+          <fieldset>
+            <legend>Join Our Mailing List</legend>
+            <div class="form-content">
+              <div class="label-input">
+                <label for="email">
+                  <span>Email <span class="required">*</span> </span></label
```
LucyMac: You shouldn't need the span here, just put the text directly in the `label`. 
The `*` can be added [using CSS](http://jsfiddle.net/et8506qj/4/)

script.js
```diff
@@ -0,0 +1,30 @@
...
+    } else {
+      navbar.classList.add("navbar");
+    }
+  } else {
+    navbar.classList.remove("navbar");
+    navbar.classList.remove("sidebar");
+    hamburgerTop.style.borderColor = "#000";
+    hamburgerMiddle.style.borderColor = "#000";
+    navbar.classList.add("hide");
+    hamburgerMenu.style.gridColumn = "1/2";
```
LucyMac: Nice mobile nav 🎉 
Watch out you're adding this grid-column rule after the nav has been closed. I don't think you need it?

LucyMac (commented):
Really nice start @mahri09! You have clearly paid attention to the details and the result looks great ⭐  Just a few comments below. 
I look forward to seeing what it looks like with some media queries in there, and it would also be great to see another page - the 'order online' page is a good challenge 😉

LucyMac (commented):
Also - great that you've split out your footer and header stylesheets for re-use 👍 
Can you store some CSS variables to save the colours and re-use them across your stylesheets?

mahri09:
Really nice start @mahri09! You have clearly paid attention to the details and the result looks great ⭐ Just a few comments below.
I look forward to seeing what it looks like with some media queries in there, and it would also be great to see another page - the 'order online' page is a good challenge 😉

>@LucyMac, thank you so much, Lucy, I was waiting for your comments :)), they're very valuable for me!
I will add as soon as finish the home page and  'order online'  thank you for the suggestion:))

index.html
mahri09: Actually, I did that first, but the text location changes with the screen change, then it's better to have it this way, I'm very happy to learn that

script.js
mahri09: I will look at it , thank you!

PR #66, opened 2021-12-08
London class 8/marina almeida HTML-CSS Week 4

MarinaAlmeida20:


css/main.css
```diff
@@ -1,3 +1,733 @@
...
+  font-size: 1rem;
+  font-weight: bolder;
+  letter-spacing: 0.5rem;
+  margin-bottom: 0.4rem;
+}
+
+.hero-content h1 {
+  font-size: 2.7rem;
+  margin-bottom: 1rem;
+  margin-top: 0.7rem;
```
LucyMac, line 191: The font-family needs defining here, it's using the wrong one.

css/main.css
```diff
@@ -1,3 +1,733 @@
...
+html {
+  box-sizing: border-box;
+  color: var(--red_brown);
+  margin: 0;
+  padding: 0;
+}
+
+body {
+  background: var(--light_grey);
+  font-family: var(--body_subheading_font-family);
```
LucyMac, line 22: The browser automatically adds an 8px margin all around the body. Remove this by setting margin to 0 here - this will make your elements 'bleed off the page', meaning they will span the full width from left to right, like in the design.

css/main.css
```diff
@@ -1,3 +1,733 @@
...
+  margin: -1.8rem 0;
+}
+
+.offer-online {
+  height: 25rem;
+  background-color: var(--green_black);
+  color: var(--white);
+}
+
+.offer-online .offer-online-h1 {
```
LucyMac, line 278: Fix the font-family here

css/main.css
```diff
@@ -1,3 +1,733 @@
...
+  color: #000;
+}
+
+.form-section label {
+  padding: 1rem 0;
+  font-size: 1.3rem;
+}
+
+.form-section input {
+  padding: 1rem 0;
```
LucyMac, line 324: It'd be good to add 1rem of padding on the left and right as well so the text doesn't touch the edges.

css/main.css
```diff
@@ -1,3 +1,733 @@
...
+  padding: 0 1rem;
+}
+
+.footer-subscribe label {
+  font-size: 1.3rem;
+  margin-bottom: 0.5rem;
+}
+
+.footer-subscribe input {
+  padding: 15px 6.6rem;
```
LucyMac, line 407: This is too much padding, it makes the inner text start too far inside. 
For inputs, padding is used just to create a little space between the text and the edges, and perhaps to create some height too. But to control the width of the input, either use grid properties, or set a `width` in `%`, e.g. `width: 100%;`
If you're having trouble fitting it in, make sure it is set to `box-sizing: border-box;` as well 💡

css/main.css
```diff
@@ -1,3 +1,733 @@
...
+
+  .feature-title-btn {
+    position: absolute;
+    width: auto;
+    padding: 14px 102px;
+    z-index: 1;
+  }
+
+  .image-products {
+    padding: 0 106px;
```
LucyMac, line 604: This padding is causing trouble on desktop. You could achieve the desired effect more easily by using CSS Grid and setting a grid-gap?

css/main.css
```diff
@@ -1,3 +1,733 @@
...
+  .feature-and-image {
+    /* width: 100%; */
+    display: flex;
+    flex-direction: row;
+    flex-wrap: nowrap;
+    justify-content: space-around;
+  }
+
+  .feature-title-btn {
+    position: absolute;
```
LucyMac, line 597: You do not want to be messing with absolute positioning here, it's not necessary - lining up 3 boxes in a row can be done using the same techniques we have learned: flexbox and/or grid. See my next comment on line 604.

index.html
```diff
@@ -1,21 +1,289 @@
...
+                  This is a great space to write long text about your company and
+                  your services. You can use this space to go into a little more
+                  detail about your company. Talk about your team and what
+                  services you provide.
+                </p>
+              </div>
+              <div class="img-about">
+                <img
+                  class="img-area-about"
+                  src="/home/marina/Desktop/CodeYouFuture/responsive-grid-project/img/mobile-area-img.jpg"
```
LucyMac, line 158: This is a path to an image on **your** computer. No one else has access to `/home/marina/Desktop` therefore no image is showing up for me. 
Replace it with the **relative** path to the image inside the project folder. It should match the same folder structure on Github, and therefore should be able to access the image. Example: `./img/mobile-area-img.jpg`

index.html
```diff
@@ -1,21 +1,289 @@
...
+              </div>
+              <div class="feature-img" title="">
+                <img
+                  src="../responsive-grid-project/img/third-image.jpg"
+                  alt=""
+                />
+              </div>
+            </div>
+          </div>
+          <div class="feature">
```
LucyMac, line 135: This whole section will need to be created using CSS Grid. If you analyse how many columns there are in this layout, you can see the 'about us' text box spans half the screen, and the images next to it span a quarter each. How would you translate this using the `fr` unit? 

P.S. I recommend you rename this class as there are quite a few 'feature' classes floating around already.

css/main.css
```diff
@@ -1,3 +1,733 @@
...
+    height: auto;
+    padding: 18px 223px;
+  }
+
+  .offer-online .offer-online-h1 {
+    font-size: 74px;
+  }
+
+  .form {
+    position: absolute;
```
LucyMac, line 648: Why the position absolute? 😱 This is once again causing problems and is not needed here. 
Think about how we solved the Karma hero section - it's the same, except there is a light box around the central content... can you try and fix it? 😉

LucyMac (commented):
Good start @MarinaAlmeida20, a good structure and attempt to follow the design. There are just a few issues with layout and positioning on desktop. Avoid `position: absolute` unless you really need it - in the cases I've highlighted it can all be achieved with flexbox and/or grid. 

Try and fix these, and add another page to your site to complete this task.

css/main.css
MarinaAlmeida20: the font Fjalla One doesn't work.

PR #67, opened 2021-12-09
London class 8 - Sima Zamani - Plant Project - Week 4

zamanisima:


css/main.css
```diff
@@ -1,3 +1,112 @@
...
+}
+.collection_succulents {
+  background: url(../img/succulent.webp);
+  height: 80vh;
+  padding-top: 1em;
+  background-position: center;
+  background-repeat: no-repeat;
+  background-size: cover;
+}
+@media screen and (min-width: 481px) {
```
LucyMac, line 93: This breakpoint is too small to switch out of mobile view. Try using a higher value, somewhere between 600px and 800px (a common one is 768px which is  the width of an iPad)

css/main.css
```diff
@@ -1,3 +1,112 @@
...
+  .hero {
+    background: url(../img/hero.webp);
+    background-repeat: no-repeat;
+    background-attachment: fixed;
+    background-size: cover;
+  }
```
LucyMac, line 99: None of these rules are different from the default (mobile) rules so you don't need to declare them here. Only re-define rules which have changed for this breakpoint.

css/main.css
```diff
@@ -1,3 +1,112 @@
...
+  .collection {
+    flex-direction: row;
+    justify-content: space-around;
+    flex-direction: row;
+    justify-content: space-around;
```
LucyMac, line 104: Remove duplicates

pages/index.html
```diff
@@ -0,0 +1,162 @@
...
+          <div class="hero_header_logo">
+            <img src="../icons/logo.svg" alt="main logo" />
+          </div>
+          <div class="hero_header_text">2 0 2 3</div>
+        </section>
+        <h1>PRICKLES & CO.</h1>
+        <h3>
+          BRING NATURE <br />
+          INDOORS
+        </h3>
```
LucyMac, line 62: Avoid forcing the line break this way. Best to let this happen naturally. You can also use left and right padding to tighten the space a little.

pages/index.html
```diff
@@ -0,0 +1,162 @@
...
+        <button class="btn">Shop Now</button>
+      </article>
+
+      <article class="collection">
+        <section class="collection_cacti">
+          <div class="Collection_item1">
+            <h1>CACTI</h1>
```
LucyMac, line 69: Probably best to use a smaller heading here, like h3 or h4

pages/index.html
```diff
@@ -0,0 +1,162 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <link rel="stylesheet" href="../css/common.css" />
+    <link rel="stylesheet" href="../css/header.css" />
+    <link rel="stylesheet" href="../css/main.css" />
+    <link rel="stylesheet" href="../css/footer.css" />
+
```
LucyMac, line 11: I notice you aren't importing any fonts here. Try importing the 2 suggested fonts (see [ReadMe](https://github.com/CodeYourFuture/responsive-grid-project#fonts)) from Google Fonts to make your page match the design.

pages/index.html
```diff
@@ -0,0 +1,162 @@
...
+        </section>
+      </article>
+    </main>
+    <footer class="footer">
+      <section class="footer_info">
+        <div class="info_item">
+          <h1>OUR STORE</h1>
+          <ul>
+            <li>Address: 500 Terry Francois Street San</li>
+            <li>Francois,CA 94158</li>
```
LucyMac, line 98: These probably shouldn't be 2 separate list items. Are they 2 items in a list, or is it just one item split onto 2 lines? 🤔

pages/index.html
```diff
@@ -0,0 +1,162 @@
...
+
+      <section class="copy_right">
+        <p>
+          © 2023 by Prickles & Co. Proudly
+        </p>
+        <p>created with <a href="https://wix.com">Wix.com</a></p>
+      </section>
+    </footer>
+  </body>
+</html>
```
LucyMac, line 162: Really nice HTML structure @zamanisima 🏅 Semantic, organised and very easy to follow.

css/main.css
LucyMac: Even better, use 2 breakpoints, so you can adapt the design for 3 screen types: mobile, tablet and desktop

css/main.css
```diff
@@ -1,3 +1,112 @@
...
+  width: 100%;
+  max-width: 100%;
+}
+.hero_header {
+  width: 100%;
+  display: flex;
+  align-items: center;
+}
+.hero h1 {
+  font-size: 8vw;
```
LucyMac, line 34: Using `vw` for your font sizes is resulting in your fonts being tiny for mobile, and too big for desktop. It's best to use fixed font sizes (using `rem` or `px`) and change this size in your media queries to give different values for different breakpoints.

LucyMac (commented):
Nice start @zamanisima. I love how you've split your CSS into separate files ⭐ 
There's a bit more work to do on the responsiveness of your page - the font sizes aren't right (too small or too big, see my comment about not using `vw` for this) and the page doesn't adapt very well to the screen size. 
Also, import some fonts to try and match the design more! 😉 

Looking forward to seeing another page too 🙂

PR #68, opened 2021-12-10
London Class 8 - Matilda Ako - HTML/CSS - Week 4

MatildaAko:
I found this a fun project to work on, it wasn't hard for me personally so I decided to practice adding some JavaScript functionalities which ended up being simpler than I thought even though it did not always behave as I expected.
This project can be viewed here: https://cyf-matildaako-plants.netlify.app/


MatildaAko:
Hi @haack My mobile version is ready for review

faq.html
```diff
@@ -0,0 +1,239 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <!-- Add a link to your custom CSS here -->
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link href="https://fonts.googleapis.com/css2?family=Fjalla+One&family=Outfit:wght@200;300;400&display=swap" rel="stylesheet" />
+    <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.15.4/css/all.css" integrity="sha384-DyZ88mC6Up2uqS4h/KRgHuoeGwBcD4Ng9SiP4dIRy0EXTlnuz47vAwmeGwVChigm" crossorigin="anonymous" />
+    <link rel="stylesheet" href="./css/variables.css" />
+    <link rel="stylesheet" href="./css/main.css" />
+    <link rel="stylesheet" href="/css/fonts.css" />
```
LucyMac: There's a dot missing here which breaks the path locally for all pages 😞

LucyMac (commented):
Awesome work @MatildaAko. Your site looks impressively close to the design, your code is neat and a lot of attention has gone into making the structure clear and semantic, as well as re-using well-named CSS classes throughout. Also great work adding some JS in there ⭐ 

Go ahead and make your pages responsive now 😃

faq.html
MatildaAko: Thank you for having a look at my work @LucyMac I have been working on the responsive design locally and have just pushed the changes I made, I still have a few more pages to complete. I also did notice the missing dot and had fixed it already, I should have done a push a long time ago.

css/layout.css
```diff
@@ -143,4 +144,93 @@ fieldset {
...
+
+.offer {
+    grid-area: of;
+}
+
+@media screen and (min-width: 980px) {
+    .white-background {
+        grid-template-columns: repeat(11, 1fr) auto;
+        grid-template-areas: "s s s s s s s s l l l b";
+    }
```
LucyMac, line 185: This isn't incorrect, but do you need to create a new grid here? Could you have set these template areas on the `.header` grid directly?

css/desktop.css
```diff
@@ -0,0 +1,126 @@
...
+    }
+    .bag {
+        justify-self: end;
+        margin-left: 36px;
+    }
+    .hero {
+        background-attachment: fixed;
+        background-position: bottom;
+        height: 100vh;
+        padding: 180px 305px 265px;
```
LucyMac: I think the padding is a little bit off here, forcing the main heading to split onto 2 lines when it could otherwise fit.

css/desktop.css
```diff
@@ -0,0 +1,126 @@
...
+    .logo-details {
+        justify-content: center;
+    }
+    .logo {
+        margin: 0 20px;
+    }
+    .product-links {
+        margin: 50px 0;
+        width: calc((100vw - 200px) / 3);
+    }
```
LucyMac: This section looks great 👌 
Nice use of `calc()` - bear in mind it might have been easier to use grid here, split into 3 columns of 1fr each with a column-gap and you're done! 🪄

css/desktop.css
```diff
@@ -0,0 +1,126 @@
...
+    .woman {
+        height: 100%;
+    }
+    .offer {
+        padding: 130px 150px 200px;
+    }
+    .about {
+        padding: 85px 95px;
+    }
```
LucyMac: I think both of these blocks could do with less left and right padding - but otherwise this whole 'about-wrapper' section looks really good. Great use of the grid here.

css/desktop.css
```diff
@@ -0,0 +1,126 @@
...
+        background: url("../images/last-plant.png");
+        background-attachment: fixed;
+        background-position: top;
+        background-size: cover;
+        padding: 100px 0;
+    }
+    .contact-form {
+        background: var(--transparentGrey);
+        width: 60%;
+    }
```
LucyMac, line 128: Love this - the fixed background, the form... all looks and feels just like intended. Nice work.

LucyMac (commented):
🙌  excellent work! You've really understood how to use the grid and have demonstrated a great eye for detail.

(still need to fix those relative paths for fonts to work)

css/desktop.css
MatildaAko: Thanks, I have used this method for a more recent grid but I will make the changes to this and any other

css/desktop.css
MatildaAko: I will change this, I also think the font size is too large so I will make these changes

css/layout.css
MatildaAko: I believed I tried it that way and had problems aligning everything and also giving the top half of the header a white background but I will try it again until I can figure it out.

css/desktop.css
MatildaAko: Thank you

css/desktop.css
MatildaAko: Thanks, from your comments, it looks like my paddings are too large so I will go back and find a different way to create the look that's shown and possibly use rem or % instead so it can scale along with the screen.

PR #72, opened 2021-12-10
Deago_Browne-LDN_CLass_8-HTML/CSS_Week_4

DeagoBrowne:
Not enough time to finish second page, so submitting as is.


css/main.css
```diff
@@ -1,3 +1,394 @@
...
-/* Dont' forget to link this file to your HTML in the <head> */
+/* Don't forget to link this file to your HTML in the <head> */
+
+:root {
+  --dBeige: #e6dacd;
+  --lBeige: #f4ece6;
+  --rBlue: #0150fd;
+  --black: #000;
+  --white: #fff;
+  --menu-speed: 0.75s;
```
mahsa2, line 11: Nice job on using variables 🌟

mahsa2: One thing that can be improved as an extra trip here is that we want to stick with a same variable nameing. Either all variable names should use `-` in places where we need space, or use camelCase for all. `-` use might be more common I think.

index.html
```diff
@@ -5,17 +5,108 @@
...
+    </footer>
 
     <!-- Add your markup here -->
 
 
     <!-- Please leave this in to credit the creators of this design -->
-    <p>This website design was created by Wix.com, and is used here for strictly educational purposes.</p>
-</body>
+    <!-- <p>This website design was created by Wix.com, and is used here for strictly educational purposes.</p>
+</body> -->
```
mahsa2, line 110: We need to keep this to respect the credit purposes.
Also, `</body>` needs a closing tag.

index.html
```diff
@@ -5,17 +5,108 @@
...
+            <div class="white-box">
+
+            </div>
+            <div class="profile">
+                <div class="p-pic">
+                    <img src="https://static.wixstatic.com/media/2e2a49_bd3a1599bc1b4265a01b2754b99b8536~mv2_d_5760_3840_s_4_2.jpeg/v1/crop/x_2703,y_43,w_2199,h_2205/fill/w_205,h_205,al_c,q_80,usm_0.66_1.00_0.01/AdobeStock_230407429.webp"
+                    alt="" class="pic"/>
+                </div>
+                <div class="p-name">
+                <h2>Maya<br>Nelson</h2>
```
mahsa2, line 53: It might be better to use css (e.g. width) to limit the space this `h2` element has to produce the new line. But this is not wrong, but we might not want this break in other screen sizes (e.g. desktop).

Also this needs one more indentation.

resume.html
```diff
@@ -0,0 +1,78 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <!-- Add a link to your custom CSS here -->
+    <link rel="stylesheet" href="./css/main.css">
+    <link rel="stylesheet" href="./css/style2.css">
```
mahsa2, line 9: I'm happy that you reused `main.css` here which is great 👏 

You can make it even better by putting the common css rules in a file (e.g. `common.css`) and include this file on all the pages it needs.
Usually this file contains the css of header, footer, and any repeated elements/css.

As a result, the browser only need to download the reapeated css not all the `main.css` files. This will speed up the website load, increases SEO ranking, and is a good practice :)

index.html
```diff
@@ -5,17 +5,108 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
     <!-- Add a link to your custom CSS here -->
+    <link rel="stylesheet" href="./css/main.css">
+    <link rel="preconnect" href="https://fonts.googleapis.com">
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+    <link href="https://fonts.googleapis.com/css2?family=Poppins:wght@600&family=Quicksand:wght@300&display=swap" rel="stylesheet">
```
mahsa2, line 11: `link` needs to be self-closed -> `<link ... />`

index.html
```diff
@@ -5,17 +5,108 @@
...
 
     <title>Responsive grid project</title>
 </head>
 
 <body>
+<header>
```
mahsa2, line 17: `<header>` needs an indentation.

index.html
```diff
@@ -5,17 +5,108 @@
...
+    
+    <div class="menu-wrap">
+        <input type="checkbox" class="toggler">
+        <div class="hamburger"><div></div></div>
+        <div class="menu">
+            <div>
+                <div>
+                    <ul>
+                        <li><a href="#">ABOUT ME</a></li>
+                        <li><a href="#">RESUME</a></li>
```
mahsa2, line 32: Here you can link to the resume page you've made.

index.html
```diff
@@ -5,17 +5,108 @@
...
+                    <img src="./img/linkedin-in-brands.svg" alt="" class="f-img"/>
+                    <img src="./img/instagram-brands.svg" alt="" class="f-img"/>
+                </div>
+            </div>
+            <div class="beige-box">
+                <div class="bBox-text">
+                <h1>Hello</h1>
+                <br>
+                <br>
+                <p>Meow</p>
```
mahsa2, line 71: 😹

mahsa2 (commented):
Pretty chuffed with the progress you've made in this assignment. 👏 
Very nice indentations and clean code 💯 🎖️ 

I know that you said at the time of this assignment, you were not familiar with `grid`, but I'm sure I'm going to see beautiful implementation of the `grid` in the future assignments :)

It's good to be able to design your page both for phones and desktop, so that you get familiar with how to use the css that minimizes the use of media and makes them work for both desktop and phone. We can talk about this on our meeting.

PR #75, opened 2021-12-10
London class 8/margarita glushkova/grid_project_juice_bar

RitaGlushkova:
Only one page is in the process - the landing page. The Mobile version looks similar, the desktop version is not complete. Still need to do the iPad version.
Still have to do Menu Page - all media Q.


css/common.css
```diff
@@ -0,0 +1,311 @@
...
+}
+.mar_bot {
+  margin-bottom: 30px;
+}
+.pad_bot {
+  padding-bottom: 24px;
+}
+.pad_top {
+  padding-top: 33px;
+}
```
LucyMac, line 23: This is fine. You could also save these values inside variables, and apply them to other classes when needed, e.g.
```
.some-other-element {
     padding: var(--space-small);
}
```

css/common.css
```diff
@@ -0,0 +1,311 @@
...
+  position: fixed;
+  width: 100%;
+  background-color: var(--white);
+  display: grid;
+  grid-template-columns: 150px auto 150px;
+  grid-template-rows: repeat(2, 44px);
+  grid-template-areas:
+    "order order order"
+    "bar . hamburger";
+  align-items: center;
```
LucyMac, line 120: You might want to add a `z-index` here to make sure the header doesn't slide underneath the 'follow' section at the bottom.

index.html
```diff
@@ -1,21 +1,331 @@
...
+          Fuel Your Body. Balance Your Mind.
+        </h1>
+        <p class="font_2 text_center width_adjust margin_auto">
+          Organic Cold Pressed Juices & Smoothies
+        </p>
+        <!--Link Explore-->
+        <a href="#" class="button_black flex_center margin_auto no_underline"
+          >Explore Flavors</a
+        >
+      </section>
```
LucyMac, line 53: This doesn't matter here, but just to clarify what a 'hero' is - I would argue that this section is the hero, _as well as_ the image below it.

index.html
```diff
@@ -1,21 +1,331 @@
...
+              <p class="font_3">
+            I'm a paragraph. Click here to add your own text and edit me. Let
+            your users get to know you.
+              </p>
+          </section>
+          <section class="feature flex_center flex_col">
+            <div class="feature-img">
+              <img src="img/Nat_ingr.svg" />
+            </div>
+            <h3 class="font_2">100% Natural Ingredients</h3>
```
LucyMac, line 89: These `h3` (like all other `h1-6` headings) have default margin set on them by the browser. You'll need to override this by setting your own margin (at the moment there's too much space between this heading and the paragraph below it)

css/common.css
```diff
@@ -0,0 +1,311 @@
...
+}
+
+.button_tr {
+  font-family: var(--buttons-font);
+  color: var(--black);
+  font-size: 1rem;
+  width: 160px;
+  height: 42px;
+  border: solid 2px var(--black);
+  background-color: none;
```
LucyMac, line 104: Really like your buttons 💅 

2 things: 
1) Avoid giving buttons a set width as it stops them from growing when their content changes. They should be able to grow / shrink with their content. The way to make them look consistent is by setting padding, but no width. This will ensure there's always the same spacing around the word(s). The same applies to height. 
2) `background-color: none` shows up as invalid in the inspector. You can set the color to `transparent` instead.

css/index_style.css
```diff
@@ -0,0 +1,206 @@
...
+  width: 100%;
+}
+.img_kale img {
+  transform: rotate(17.142845342601447deg);
+  width: 100%;
+}
+.follow img {
+  width: 100%;
+}
+@media screen and (min-width: 481px) {
```
LucyMac, line 128: 481px is a little too early to be switching out of mobile view. I'd keep mobile view until 600px minimum at this stage. 
Where you decide to 'break' will depend on how many breakpoints you have (i.e. 1 breakpoint = only 2 versions, therefore better to break later. Whereas 4 breakpoints = 5 versions and therefore you'll spread the breakpoints out more).

css/index_style.css
```diff
@@ -0,0 +1,206 @@
...
+  .vision div {
+    width: 50%;
+  }
+  .vision .flex_col {
+    align-items: flex-start;
+    text-align: left;
+    width: 490px;
+  }
+  .vision img {
+    width: 557px;
```
LucyMac, line 167: You're setting quite a few fixed widths which I wouldn't recommend. These mean your layouts will look perfect at a very particular screen size, but won't adapt to for example a slightly larger phone, or a slightly smaller desktop. 
It's best to set widths either in `%` or in `fr` if using grid. This ensures you are setting proportions instead of fixed dimensions. In the case of this image, it's basically taking up half of the screen in desktop view, so that's what you should be aiming for rather than a fixed width.

index.html
```diff
@@ -1,21 +1,331 @@
...
+              I'm a paragraph. Click here to add your own text and edit me.
+            </p>
+          </section>
+        </div>
+      </section>
+
+      <!--Section Follow Us-->
+      <section class="text_center">
+        <h2 class="font_1">Follow Us</h2>
+        <section class="follow">
```
LucyMac, line 252: This section looks awesome, well done 🌟

LucyMac (commented):
Really great work on the mobile view @RitaGlushkova, it looks awesome. Great HTML structure and spotless styling ✨ I've left a few comments about keeping the styling more fluid by default rather than fixing it to the current screen size.
Looking forward to seeing it fully responsive 😃

RitaGlushkova:
@LucyMac Hi Lucy thank you very much for taking your time to review my work. Your feedback is very helpful. I am planning to finish this project by the end of January. I will send you an update once it is done :)

PR #76, opened 2021-12-10
london class8-Laleh Koupaei-Html/Css-Week4

lalehkoupa:
I have finished my main plant page (mobile version)


Simon2828 (commented):
Great work again Laleh! Your page looks very good - your HTML is semantic and you've used CSS grid and media queries 😁 👍

PR #78, opened 2021-12-10
London class 8/berkeli halmyradov

berkeli:
@gixo
Final project for HTML/CSS.
I have created the main page (about us, contact) and shop page.


css/common.css
```diff
@@ -0,0 +1,83 @@
...
+  background-color: var(--red-light);
+}
+.underline {
+  color: var(--red-brown);
+  font-size: 4rem;
+  font-weight: 600;
+  line-height: 0;
+}
+.hero {
+  width: calc(100% + 4rem);
```
LucyMac: Why the larger-than-screen size?

css/common.css
```diff
@@ -0,0 +1,83 @@
...
+  background-repeat: no-repeat;
+  background-size: cover;
+  margin: 0 -2rem;
+  display: flex;
+  align-items: center;
+  flex-direction: column;
+  background-color: rgba(0, 0, 0, 600);
+  box-shadow: inset 0 0 0 2000px rgba(0, 0, 0, 0.25);
+}
+.hero > * {
```
LucyMac: Works here, but generally best to avoid the `*` selector, other than setting very basic rules for the whole page (like `margin: 0` etc)

css/common.css
```diff
@@ -0,0 +1,83 @@
...
+.hero h1 {
+  font-size: 4rem;
+  font-family: var(--heading-font);
+  font-weight: 400;
+  margin-bottom: 2rem;
+}
+@media screen and (min-width: 900px) {
+  main {
+    margin-top: 15vh;
+  }
```
LucyMac: Since you're (I think) wanting to set `main` to start where the `header` ends, it would make sense to give it the same margin-top as the height of the header. So if the header is 100px high, you'd give main `margin-top: 100px`. 
To help with keeping track of this, you could even save this value inside a variable, and use it for the header height and the main margin-top. Like this, if you decide to change it, you can update the variable and it will change in both places and still work ✨

css/main.css
```diff
@@ -1,3 +1,171 @@
...
+    height: 60vh;
+    overflow: hidden;
+  }
+  .product img {
+    height: 100%;
+    width: auto;
+  }
+  .about-us {
+    grid-template-rows: 1fr 1.3fr;
+    height: 150vh;
```
LucyMac: Best not to set fixed heights unless you really need to. In this case, I don't think you do, and it's breaking the contact form layout a little at the start of your desktop view (between 900px and 1160px approx)

LucyMac: Other than that, great use of Grid on this section 👌

LucyMac (commented):
Excellent work @berkeli - you've followed the designs with a lot of eye for detail and made an effort to keep things very organised. 
In the future, try and use more than one breakpoint so your design adapts even better for tablet e.g. products can be 2 per row instead of 1.
I've left a couple of small comments but otherwise very well done 👏

PR #79, opened 2021-12-11
Ldn8 / Final HTML-CSS project / Saman Zahedi

SamanZahedi:


css/main.css
```diff
@@ -1,3 +1,222 @@
...
+.profile-img {
+  display: block;
+  border-radius: 50%;
+  width: 44%;
+  margin: 15px auto 18px auto;
+}
+
+header > p:nth-child(1) {
+  margin: 0;
+  display: grid;
```
beckyf0keefe: This is the same as last week, I'm guessing you wrote this before you saw my feedback, but just remember you don't need to use display: grid on the child, only the parent

css/main.css
```diff
@@ -1,3 +1,222 @@
...
+:root {
+  --dark-beige: #e6dacd;
+  --light-beige: #f4ece6;
+  --royal-blue: #0150fd;
+  --black: #000;
+  --white: #fff;
+}
+
+body {
+  width: 320px;
```
beckyf0keefe: It is best never to set the width on the body, in this case as we are doing mobile first, you don't have to set a width at all, just adding one for the width with the media query for bigger screens will be sufficient

css/main.css
```diff
@@ -1,3 +1,222 @@
...
+  line-height: 2em;
+  letter-spacing: 0.02em;
+  font-size: 14px;
+}
+h1 {
+  font-family: "Poppins", sans-serif;
+}
+.hero {
+  position: relative;
+  background-color: rgb(248, 234, 234);
```
beckyf0keefe: Remember to apply the colour variables that you defined at the top

css/main.css
```diff
@@ -1,3 +1,222 @@
...
+  background-color: var(--royal-blue);
+  color: var(--white);
+  font-weight: bold;
+}
+main .icons {
+  width: 100%;
+  margin-bottom: 10px;
+  background-color: white;
+  justify-content: center;
+  position: absolute;
```
beckyf0keefe: Can you think of a way to do this without using position: absolute?

css/main.css
```diff
@@ -1,3 +1,222 @@
...
+.intro {
+  background-color: burlywood;
+  position: absolute;
+  z-index: -1;
+  top: 50px;
+  padding: 260px 28px 21px 28px;
+}
+
+/* footer style */
+footer {
```
beckyf0keefe, line 152: Footer is looking really good, good use of grid here, just remember that the child doesn't need to have display: grid

css/main.css
```diff
@@ -1,3 +1,222 @@
...
+.icons img {
+  padding: 8px;
+}
+/* intro style */
+.intro {
+  background-color: burlywood;
+  position: absolute;
+  z-index: -1;
+  top: 50px;
+  padding: 260px 28px 21px 28px;
```
beckyf0keefe, line 148: How can we do this without the padding? Try changing your grid slightly so that you don't need so much padding, large amounts of padding will often cause issues later down the line and make your site less reactive.

css/main.css
```diff
@@ -1,3 +1,222 @@
...
+  font-size: 14px;
+}
+h1 {
+  font-family: "Poppins", sans-serif;
+}
+.hero {
+  position: relative;
+  background-color: rgb(248, 234, 234);
+  margin-left: 6%;
+  margin-right: 2%;
```
beckyf0keefe, line 93: How could we centre the hero without using margin left and right? By setting a set width of the container and centring it it would be more reactive? This can be a challenge to do before next week

css/main.css
```diff
@@ -1,3 +1,222 @@
...
 
+/* Add your CSS here */
 /* Dont' forget to link this file to your HTML in the <head> */
+:root {
+  --dark-beige: #e6dacd;
+  --light-beige: #f4ece6;
+  --royal-blue: #0150fd;
+  --black: #000;
+  --white: #fff;
+}
```
beckyf0keefe: Good to see you creating variables

beckyf0keefe (commented):
Good start, glad to see you worked out a way to get the hovering card effect, I have added a few suggestions of some changes that would help make your app more reactive which would be great to see if you have time this week

PR #80, opened 2021-12-11
London-class-8-Responsive-Grid-Project-JuiceBar-BusraErdogan

BusraElmasErdogan:
… file for Mobile

Juice bar website project mobile version completed.


css/main.css
```diff
@@ -1,3 +1,87 @@
...
+  --color_17: 212, 247, 147;
+  --color_18: 94, 125, 31;
+  --color_19: 80, 101, 64;
+  --color_20: 57, 75, 42;
+  --color_21: 245, 232, 255;
+  --color_22: 235, 203, 247;
+  --color_23: 192, 110, 255;
+  --color_24: 126, 55, 189;
+  --color_25: 71, 32, 105;
+  --color_26: 252, 185, 146;
```
LucyMac, line 28: I don't recommend this naming. Very hard to use, no idea what's what. Plus you won't use most of these colours. 
It would be better practice to name them descriptively (and also more common to use the hex value. Don't copy what wix have done, they have terrible practices 😄)

index.html
```diff
@@ -1,21 +1,269 @@
...
 <html lang="en">
-<head>
-    <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <meta http-equiv="X-UA-Compatible" content="ie=edge">
-    <!-- Add a link to your custom CSS here -->
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
```
LucyMac, line 7: You will need to import your fonts in here for them to work. Questrial etc are not native system fonts, so they need to be added to the project. You'll find them in Google Fonts along with the documentation of how to import them.

index.html
```diff
@@ -1,21 +1,269 @@
...
+    <main>
+      <section class="explore-flavors">
+        <div class="first">
+          <h1>Fuel Your Body. Balance Your Mind.</h1>
+        </div>
+        <div class="second">
+          <h2>Organic Cold Pressed Juices & Smoothies</h2>
+        </div>
+        <div class="explore-button">
+          <button class="explore-flavors-button">Explore Flavors</button>
```
LucyMac, line 48: This button could occur elsewhere in the website. Give it a more generic name, so you can re-use it if needed. 
A common class name for buttons is `button--primary` and `button--secondary` but you could also call them `button--dark` and `button--outline` (if you're wondering about the double dash `--`, this follows the [BEM notation](https://css-tricks.com/bem-101/))

index.html
```diff
@@ -1,21 +1,269 @@
...
-</html>
\ No newline at end of file
+    <main>
+      <section class="explore-flavors">
+        <div class="first">
+          <h1>Fuel Your Body. Balance Your Mind.</h1>
+        </div>
+        <div class="second">
+          <h2>Organic Cold Pressed Juices & Smoothies</h2>
+        </div>
```
LucyMac, line 46: I'm not sure you need to wrap the text inside divs here

css/choose-your-path.css
```diff
@@ -0,0 +1,52 @@
...
+  text-align: center;
+}
+
+.path h2 {
+  width: 200px;
+  font-size: 22px;
+}
+
+.path p {
+  width: 255px;
```
LucyMac, line 41: You are giving a lot of elements a fixed width. This goes against the 'responsive' principles. If my mobile phone is bigger, I expect the text to spread out more so it makes use of my bigger screen. You have made everything look perfect but at a very specific mobile screen size - this should be more adaptable. 
Things that should stay fixed are for example the spacing on the sides of the content: there are 20px of empty space on the left and right of the content. This should stay the same regardless of the size of the mobile phone. Apart from those 20px margins, the content should be able spread out more, where appropriate.

LucyMac: This is why we use `%` for widths , or `fr` for grid items - because they are relative / proportionate and they adapt to the size of the screen. 
Fixed widths using pixels or rems will not adapt.

css/juice-bar-green.css
```diff
@@ -0,0 +1,34 @@
+.green-part {
+  background-color: rgb(var(--color_16));
```
LucyMac, line 2: You can add padding here (1 or 1.5rem) and then let all the content inside be at 100% width.

css/choose-your-path.css
```diff
@@ -0,0 +1,52 @@
...
+.path {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  text-align: center;
+}
+
+.path h2 {
+  width: 200px;
+  font-size: 22px;
```
LucyMac, line 37: Try and keep vertical spacing consistent with the design. The gap here is much bigger than in the design. This is because the browser has default margins on `h2` and `p`, creating a larger space between the heading and the paragraph. You need to overwrite these by setting your own margins. 

As a general rule, it's easier to only give elements spacing on the **bottom** - not on the bottom AND the top. This is because it's easier to keep track of. 
So you should set `margin-top: 0` for all the elements, and only give them a `margin-bottom`.

css/choose-your-path.css
```diff
@@ -0,0 +1,52 @@
...
+  font-size: 15px;
+}
+
+.view-all {
+  background-color: rgb(var(--color_21));
+  padding: 20px 55px;
+  margin-top: 30px;
+  margin-bottom: 30px;
+  border: solid;
+  font-size: larger;
```
LucyMac, line 51: It's not good practice to use this vague terminology. What size is larger? Prefer `rem` for font sizes.
(In this particular case, the button font size seems the same as the paragraph size in the design, don't you think?)

LucyMac (commented):
Nice start @BusraElmasErdogan! 

My main comments would be:
- to avoid setting fixed widths to elements as it stops them from growing for larger screens. See my comments below - content should adapt to the device it's being viewed on. 
- to be careful not to duplicate CSS. You have split your CSS so much that you are repeating the same rules in several stylesheets. Remember to put styles that occur several times together and re-use them. 

Add some fonts to make it look like the design, and some media queries to change the layout for tablet and desktop ✏️

PR #84, opened 2021-12-11
London class 8/timea reich/responsive grid project

TimeaReich:


css/main.css
```diff
@@ -1,3 +1,769 @@
...
+}
+.background-color-1-section  > h2{
+    grid-column: span 12;
+}
+.background-color-1-section  > h3{
+    grid-column: span 12;
+}
+.background-color-1-section  > p{
+    grid-column: span 12;
+}
```
LucyMac: These rules are still 'active' in desktop view, and are breaking the grid. You have attempted to override this with the `..color-1-section-img-1-header` class, however this is [less specific](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) than `..background-color-1-section > p` (one class + one element 'wins' over just one class). 

To avoid this kind of problem, don't target the same element in different ways. Choose a class for your element and stick to it, it will avoid unexpected behaviour like this.

css/main.css
```diff
@@ -1,3 +1,769 @@
...
+       background-color: var(--main-color);
+   }
+   .background-color-1-section-img-1 {
+       grid-column: 1/5;
+       grid-row: 3/6;
+   }
+   .background-color-1-section-img-2 {
+        grid-column: 5/9;
+        grid-row: 2/5;
+   }
```
LucyMac, line 617: You've used the grid in a clever way here. I think you could have made your life easier though by wrapping each (image + heading + paragraph) into a `div`, so you would only have to place each div on the grid and the contents would move accordingly. The way you've done it, you're moving each individual item which is much harder to keep track of.

css/main.css
```diff
@@ -1,3 +1,769 @@
...
+        grid-column: 8/11 ;
+        grid-row: 3/5; 
+        width: 250px;
+        height: 300px;  
+    }
+    .follow-section-img-4 {
+        grid-column: 11/13 ; 
+        width: 250px;
+        height: 300px;  
+    }
```
LucyMac: Remember when styling elements that have a lot in common like these 4 images that it's good to group the styles in common into one class and re-use that, and then add on a separate class for each image with the styles that are different.
`.slanty-image {
        width: 250px;
        height: 300px;  
}`
`.img--pointing-left {
       transform: rotate(-20deg);
}`
`.img--pointing-right {
       transform: rotate(20deg);
}`
and then
`<div class="slanty-image img--pointing-left"></div>`
`<div class="slanty-image img--pointing-right"></div>`
`<div class="slanty-image img--pointing-left"></div>`
`<div class="slanty-image img--pointing-right"></div>`

index.html
```diff
@@ -4,18 +4,172 @@
...
+    </section>
+
+
+    <section class="background-main-color-section">
+        <div class="background-main-color-section-img"></div>
+        <h2>Discover Our Special Cleanse Kit</h2>
+        <p>I'm a paragraph. Click here to add your own text and edit me. It’s easy. Just click “Edit Text” or double click me to
+        add your own content and make changes to the font. Feel free to drag and drop me anywhere you like on your page. I’m a
+        great place for you to tell a story and let your users know a little more about you.​</p>
+        <button class="button-no-hover"><a href="#">Try It</a></button>
```
LucyMac, line 104: Again here if you put these 3 elements (heading, text and button) inside a div, you'll be able to place these on the grid more easily.

css/main.css
```diff
@@ -1,3 +1,769 @@
...
+       gap: 1rem;
+   }
+   .background-main-color-section-img {
+       grid-column: 2/7;
+       grid-row: span 5;
+   }
+   .background-main-color-section > h2 {
+       grid-column: 7/13;
+       padding: 1rem 0;
+   }
```
LucyMac, line 666: I think the image could start in the first column here (1/6) and the text start from 6 to 13, to better match the design. 
Remember to give vertical breathing space as well (the design has white space at the top of this block). You can do this by adding padding-top and bottom to the whole section (easier than doing this with the grid).

css/order online.css
```diff
@@ -0,0 +1,407 @@
+
```
LucyMac, line 1: Avoid spaces in your file names. Best to name it order-online.css and same for html.

order online.html
```diff
@@ -0,0 +1,150 @@
...
+
+        <h1>Menu</h1>
+        <section class="cold-pressed">
+            <h2 class="product-header">Cold Pressed Juice</h2>
+            <div class="up-icon"></div>
+
+            <div class="cold-pressed-img1"></div>
+            <h3 class="product-name">Tangerine</h3>
+            <p class="product-price">$8</p>
+            <p class="product-description">I’m a dish description. Click “Edit Menu” to open the ...<div class="full-text">Restaurant Menu editor and change my text.</div></p>
```
LucyMac: Try to get in the habit of wrapping elements that 'go together' into a div so that they can more easily be styled, moved around, and aligned.

css/order online.css
```diff
@@ -0,0 +1,407 @@
...
+    grid-column: 11/ 12;
+    width: auto;
+    height: 100%; 
+} 
+main {
+    grid-column: span 12;
+    display: grid;
+    grid-template-columns: repeat(12, 1fr);
+    grid-template-rows: repeat(6, auto);
+}
```
LucyMac, line 114: I like what you've done with this page 🌟 
I think in this case it might be easier to split the main area into 3 columns of unequal size, and only have one row. And then inside each column, you might not need to use CSS Grid. For example, the central section can probably be achieved quite easily with flexbox: wrap each product/item block into a div, and align the image + text + price inside that div into a row using flexbox and padding. Probably easier than dealing with lots of columns.

LucyMac (commented):
Really nice work @TimeaReich, great looking page and really good attempts to use CSS Grid. I think in some cases perhaps you've over-used it and made your life a little harder 😉 
Remember you can also use flexbox when it makes sense, and grouping elements together into containers can also make it much easier to move these blocks around. 
Great work though, keep going! ⭐

PR #85, opened 2021-12-11
london-class 8/responsive-grid-project/hamra-khan

HamraKhan:


css/faq.css
```diff
@@ -0,0 +1,144 @@
+:root {
+    --red-brown: #A05941;
```
ewintram: awesome to see your use of CSS variables! I think you only need to define them in the one CSS file that you import into all pages, i.e. main.css

css/faq.css
```diff
@@ -0,0 +1,144 @@
...
+
+body {
+    background-color: var(--light-grey);
+    /* font-family: 'Fjalla One', sans-serif;
+    font-family: 'Outfit', sans-serif; */
+    display: grid;
+    margin: 0;
+    grid-template-columns: repeat(4 1fr);
+    grid-template-rows: auto;
+    grid-template-areas:
```
ewintram, line 7: great use of all these CSS grid properties! I think it would make sense to define styles on the `body` in a CSS file that is shared across all pages (rather than in this page-specific file), as they should probably be consistent across the site.

css/form.css
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.field-title {
+    font-family: 'Outfit', sans-serif;
+    padding-bottom: 0.8em;
+}
+
+.form input[type=text], 
+     [type=email], 
+     [type=tel],
```
ewintram: well done targeting the different input types

css/header.css
```diff
@@ -0,0 +1,258 @@
...
+    filter: invert(38%) sepia(12%) 
+    saturate(2170%) hue-rotate(329deg) 
+    brightness(100%) contrast(88%);
+}
+
+/* LOGO & NAVBAR */
+.header-nav {
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
```
ewintram: great use of flex-box

css/header.css
```diff
@@ -0,0 +1,258 @@
...
+    margin: 11px 0;
+    transition: all .2s ease-in-out;
+}
+
+.nav-icon:hover:before {
+    transform: translateY(12px) rotate(135deg);
+}
+
+.nav-icon:hover:after {
+    transform: translateY(-12px) rotate(-135deg);
```
ewintram, line 119: wow, well done achieving this transform effect!

css/header.css
```diff
@@ -0,0 +1,258 @@
...
+}
+
+.nav {
+    visibility: hidden;
+}
+
+/* MEDIA QUERIES */
+
+/* tablet */
+@media screen and (min-width: 900px) {
```
ewintram, line 142: good job building this mobile-first. It makes sense to keep media query styles close to the base styles they're overriding, so the `body` styles here might be better placed in main.css

css/main.css
```diff
@@ -1,3 +1,309 @@
...
+
+/* HERO SECTION */
+.hero {
+    background-image: url("../images/prickles-hero-img.png");
+    grid-area: hero;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    background-size: cover;
+    background-repeat: no-repeat;
```
ewintram, line 31: perfect use of a background image

faq.html
```diff
@@ -0,0 +1,216 @@
+<!DOCTYPE html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <meta http-equiv="X-UA-Compatible" content="ie=edge">
+        <title>Responsive grid project</title>
+        <link rel="stylesheet" href="./css/header.css"/>
+        <link rel="stylesheet" href="./css/faq.css"/>
+        <link rel="stylesheet" href="./css/footer.css"/>
```
ewintram, line 11: I really like how you break down the CSS files as you are starting to think in terms of components

faq.html
```diff
@@ -0,0 +1,216 @@
...
+        <link href="https://fonts.googleapis.com/css2?family=Fjalla+One&family=Outfit:wght@200&display=swap" rel="stylesheet">
+        
+    </head>
+
+    <body>
+        <header class="header">
+            <div class="header-top-half">
+                <!-- SEARCH BAR -->
+                <div class="search-container">
+                    <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet"/>
```
ewintram: I think `link` tags should be in the `head` element

faq.html
```diff
@@ -0,0 +1,216 @@
...
+                    <i class="fa fa-search" aria-hidden="true" ></i>
```
ewintram, line 22: nice us of FA icons!

faq.html
```diff
@@ -0,0 +1,216 @@
...
+        <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+        <link href="https://fonts.googleapis.com/css2?family=Fjalla+One&family=Outfit:wght@200&display=swap" rel="stylesheet">
+        
+    </head>
+
+    <body>
+        <header class="header">
+            <div class="header-top-half">
+                <!-- SEARCH BAR -->
+                <div class="search-container">
```
ewintram: This div could actually be a form. There are some other attributes you can use to ensure an input of type `search` is used in an accessible way, check out the docs https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search

faq.html
```diff
@@ -0,0 +1,216 @@
...
+                    <link href="https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css" rel="stylesheet"/>
+                    <i class="fa fa-search" aria-hidden="true" ></i>
+                    <input class="input-field" type="text" placeholder="Search..." name="search">
+                </div>
+
+                <!-- SHOPPING BASKET -->
+                <div class="shopping-basket">
+                    <img src="./icons/shopping-basket.svg" class="svg-color" alt="shopping basket icon"/>
```
ewintram: great that you're using svgs for icons

faq.html
```diff
@@ -0,0 +1,216 @@
...
+                <div class="social">
+                    <a target="_blank" href="#">
+                        <img class="social-media-icons" src="./icons/instagram-icon.svg" alt="instagram logo"/>
+                    </a>
+                </div>
+            </div>  
+            
+            <!-- MAP DISPLAY -->
+            <div class="map-container">
+                <iframe
```
ewintram, line 207: nice job adding the map

ewintram (commented):
The site looks super! Well done for tackling 2 pages and still making your code really DRY

PR #87, opened 2021-12-11
London 8/maira

mairacagri:
not happy with my page ... still a lot to do


index.html
```diff
@@ -4,14 +4,99 @@
...
+    
+</header>
+
+<main class="main">
+    
+ <section class="centercontent"> 
+    <div class="estb_2023">
+        <p class="est"><u>ESTB</u></p>
+        <div class="brand_logo">
+            <svg preserveAspectRatio="xMidYMid meet" data-bbox="10.326 50.3 180.129 101.208" xmlns="http://www.w3.org/2000/svg" width="200" height="200" viewBox="10.326 50.3 180.129 101.208" role="presentation" aria-hidden="true">
```
mahsa2, line 65: Svgs can be treated like an image. You may store the `svg` in a file and you can give it name and end it with `.svg`. Then include it in the html.
```
<img src = "happy.svg" alt="My Happy SVG"/>
```

css/main.css
```diff
@@ -1,3 +1,190 @@
...
+
+.wrapper_photo {
+  display: grid;
+  
+  gap:20px;
+}
+
+.photo img {
+  width: 100%;
+  height:auto;
```
mahsa2, line 188: `auto` is the default value of `height`, so you may decide not to write it here.

index.html
```diff
@@ -4,14 +4,99 @@
...
+          </div>
+        <p class="2023"><u>2023</u></p> 
+      </div>
+      
+      <div class="brand_name">
+         <h1>PRICKLES & CO</h1>
+         <p>BRING NATURE INDOORS</p>
+      </div>
+
+      <div class="button">
```
mahsa2, line 81: The `<button>` can be sit here without a wrapper div. Then you might need to give it `display: flex`, `display: inline-block` or `display: block` to position it on the page.

mahsa2 (commented):
It's okay that you couldn't finish. It happens. You did your best 👍 Well done for the effort you did put on it.

One thing to remeber is that always start working on the parts of the assignment that are new to you and you can learn something from. The goal of these exercises is to learn.

PR #88, opened 2021-12-11
london-class-8/NesrinSevi/final-HTML-CSS

NesrinSevi:


css/main.css
```diff
@@ -1,3 +1,276 @@
...
+  --black: #000;
+  --white: #fff;
+  --red-brown: #e95e30;
+  --light-grey: #ededee;
+  --green-black: #133032;
+  --orange-dark: #f7dca2;
+  --main-light: #8b72a8;
+  --main-dark: #69b8a4;
+  --secondary-light: #fdd8bd;
+  --secondary-light: #eecbf8d2;
```
LucyMac, line 15: duplicate colour

css/main.css
```diff
@@ -1,3 +1,276 @@
...
+  grid-column: 1 / 3;
+  grid-row: 3 / -2;
+  margin: auto;
+  padding: 10px;
+  color: var(--red-brown);
+}
+
+.main {
+  width: 100%;
+  background-size: cover;
```
LucyMac, line 92: This rule will not do anything as you have not set a background image.

css/main.css
```diff
@@ -1,3 +1,276 @@
...
+  border: none;
+  align-items: center;
+  font-size: 1.7rem;
+}
+.succulent {
+  color: var(--red-brown);
+  border: none;
+  align-items: center;
+  font-size: 1.7rem;
+}
```
LucyMac, line 168: CSS classes were invented to avoid this scenario. All 3 classes have exactly the same styles in them - create one class only, and re-use it!

css/main.css
```diff
@@ -1,3 +1,276 @@
...
+
+.welcome img {
+  position: relative;
+  margin-top: 6rem;
+}
+
+.esbt {
+  position: absolute;
+  left: 35px;
+  top: 22rem;
```
LucyMac, line 105: 😱 Don't use position absolute! You don't need it here - use flexbox to center the text inside the "welcome" section.

1) remove  `position: absolute;` and `left` and `top` rules from all the elements inside "welcome"
2) remove the welcome img - it's pushing all the content down. You need a background-image on "welcome" instead.
3) set "welcome" to `display : flex` and `flex-direction: column`

This should be exactly the same set up as the hero section of the Karma homepage: a big box with a background image, and some text elements inside which are listed as a column and centered in the middle of the box.

css/main.css
LucyMac: You also need to set all of these boxes to be flex containers, so that you can center their content (the image, the text and the button)

css/main.css
```diff
@@ -1,3 +1,276 @@
...
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  text-align: center;
+}
+
+.info {
+  margin-top: 5px;
+  padding: 5px;
+}
```
LucyMac, line 236: This is good! You are re-using the `info` class for each block, and setting flex rules on the container (`info-footer`) to center the children - good work 👌

index.html
```diff
@@ -1,21 +1,219 @@
...
+    <main class="main">
+        <article class="hero">
+        <section class="welcome">
+
+            <img src="./plant img/background-img.jpeg" alt="background-img" class="background-img">
+
+            <h3 class="esbt">E S B T</h3>
+            <div class="welcome-logo">
+            <img src="./plant img/first-icon.png" alt="first-icon" class="first-icon">
+            </div>
```
LucyMac, line 69: Please improve your indentation here, it makes it hard to read HTML if it is not properly indented.

LucyMac (commented):
Good start Nesrin but more work to do on the CSS. 
You have made a good attempt with a clear HTML structure but the styling can be improved further. Remember, don't use `position: absolute` unless you really need to. Most of the time, layout is achievable using flexbox and/or grid. In your case, using flexbox everywhere would solve a lot of issues 😉 

Keep going!

PR #90, opened 2021-12-11
London-Class-8-Delroy Gayle-HTML/CSS-Responsive Grid Project-Week 4

DelroyGayle:
Managed the mobile look.
Managed one media display
Working on the others


index.html
```diff
@@ -1,21 +1,325 @@
...
+                </div>
+            </div>
+        </header>
+
+        <figure class="OrangePict">
+            <img src="img/orange-thumbnail.webp" alt="display of orange fruit and bottles">
+        </figure>
+
+        <figure class="Desktop-OrangePict">
+            <img src="img/desktop-orange.jpeg" alt="display of orange fruit and bottles">
```
amellie: This image extends outside of the browser's maximum width. To fix this, you can apply the following styles to the image itself. The idea is to make the image covers the whole container (in this case, figure.Desktop-OrangePict, which is a grid), regardless of the width/height of the container. This is the way to make the image responsive.

```
.images {
    object-fit: cover;
    width: 100%;
    height: 100%;
    object-position: center top;
}
```

With this solution, you actually do not need to have `OrangePict` figure class anymore

css/main.css
```diff
@@ -1,3 +1,673 @@
...
+    align-items: center;
+}
+
+.OrangePictures {
+    grid-area: orange;
+}
+
+.Desktop-OrangePict {
+
+    display:none;
```
amellie: this class and `OrangePict` class have the same content (i.e. orange picture), so we only one class as explained above.

index.html
```diff
@@ -1,21 +1,325 @@
...
 
-    <title>Responsive grid project</title>
+    <title>Responsive grid project: Juice Bar</title>
 </head>
 
 <body>
+    <main class="home">
+        <header class="home-header">
+            <div>
+                <button type="button" class="blockbutton">Order Online</button>
```
amellie, line 31: In desktop version, the Order Online text should not span to all columns. It's located on the right corner of the page

index.html
```diff
@@ -1,21 +1,325 @@
...
+                <div class="pagetopflex">
+                    <p class="leftinset">Juice Bar</p>
```
amellie: Juice Bar should in the middle for the desktop version

index.html
```diff
@@ -1,21 +1,325 @@
...
+                    <!-- &#9776; is the HTML symbol for the “3 lines hamburger icon” -->
+                    <p class="rightinset">&#9776;</p>
+                </div>
+            </div>
+        </header>
+
+        <header class="home-body1">
+            <div>
+                <p class="header1format">
+                    <span>Fuel Your Body.</span>
```
amellie, line 44: Please don't use `<span>` anymore to break the words as it makes it non-responsive. Instead, you can change the size of the font or change the width of the container

index.html
```diff
@@ -1,21 +1,325 @@
...
+                    <p class="leftinset">Juice Bar</p>
+                    <!-- &#9776; is the HTML symbol for the “3 lines hamburger icon” -->
+                    <p class="rightinset">&#9776;</p>
+                </div>
+            </div>
+        </header>
+
+        <header class="home-body1">
+            <div>
+                <p class="header1format">
```
amellie: is header should be in a `<p>` tag?

css/main.css
```diff
@@ -1,3 +1,673 @@
...
+.flex-3icons {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    gap: var(--fm-1em);
+}
+
+.ThreeIcons{
+    /* 70px X 70px */
+    width:4.375em;
```
amellie: there is no need to control the width and height of the images. Removing these will still work

css/main.css
```diff
@@ -1,3 +1,673 @@
...
+
+
+    .display-3icons {
+        flex-direction: row;
+        align-items: center;
+        gap: var(--fm-1em);
+    }
+
+    .flex-3icons {
+        display: block;
```
amellie: not sure why this is a block in desktop and not keep it as a flex?

index.html
```diff
@@ -1,21 +1,325 @@
...
+                <span>Try Now</span>
+            </p>
+
+            <p class="header1format">
+                <span>What Makes Us</span>
+                <span>Different</span>
+            </p>
+        </article>
+
+        <article class="display-3icons">
```
amellie: Does this content contain an article? What would be a better tag for this?

index.html
```diff
@@ -1,21 +1,325 @@
...
+        </article>
+
+        <article class="GreenArea">
+            <figure>
+                <img src="img/green-thumbnail.webp" class="GreenPicture"
+                    alt="a display of avocado, orange & papaya against a green background">
+            </figure>
+            <div class="GreenHeader">
+                    <p class="subheader2format text">
+                        <span>Boosting</span>
```
amellie, line 131: Again, no spans here please and use the correct tag for a heading

index.html
```diff
@@ -1,21 +1,325 @@
...
+                        <span>Immunity the Way</span>
+                        <span>Nature Intended</span>
+                        <span></span>
+                    </p>
+                    <p class="paragraphtext2">
+                        <span></span>
+                    </p>
+
+                    <div class="centre">
```
amellie, line 140: I see a lot of space after this div in the desktop version

index.html
```diff
@@ -1,21 +1,325 @@
...
+                </div>
+            </div>
+        </article>
+
+        <p class="subheader2format text GetJuiceBarHeader">
+            <span>Get Juice Bar</span>
+            <span></span>
+        </p>
+
+        <article class="display-3items-flexrow">
```
amellie, line 234: Somehow, this section has a white right gap in the desktop version. The right way is to add the background color to the `display-3items-flexrow` class, instead of the `display-3numbers` class

index.html
```diff
@@ -1,21 +1,325 @@
...
+
+        <article class="FollowUs">
+            <div class="centre">
+                <a href="https://www.instagram.com/wix/" class="AButton WButton FButton">Follow Us</a>
+            </div>
+        </article>
+
+    </main>
+
+    <footer class="FooterArea">
```
amellie, line 276: Not really sure why the footer has a right white gap in the mobile version?

index.html
```diff
@@ -1,21 +1,325 @@
...
+                    <p class="subheader2format">
+                        <span>Curbside Pick Up</span>
+                    </p>
+                    <p class="paragraphtext3">
+                        <span></span>
+                    </p>
+            </div>
+        </article>
+
+        <p class="gap1"></p>
```
amellie, line 266: I would rather put apply the margin to the next tag (i.e. `<article class="FollowUs">`). For example: `<article class="FollowUs gap1">`. This is much cleaner.

css/main.css
```diff
@@ -1,3 +1,673 @@
...
+        justify-content: space-between;
+  
+    }
+
+    .FooterArea {
+        grid-area: footer; 
+        display: grid;
+        grid-template-columns: 1fr 1fr 1fr;
+        grid-template-rows:    1fr 1fr 1fr 1fr;
+        grid-template-areas:
```
amellie: I don't see most of these areas defined anywhere (or they are wrongly defined inside another container, which makes it non-grid)

css/main.css
```diff
@@ -1,3 +1,673 @@
...
+        "greenpict greenhr"
+        "greenpict greenbutton";
+        background-color: var(--light-green);
+    }
+
+
+    .row-flexdisplay {
+        display:flex;
+        flex-direction:row;
+        justify-content: space-between;
```
amellie: space-around is actually better here, so that we don't have so much space in between

css/main.css
```diff
@@ -1,3 +1,673 @@
...
+}
+
+
+/* Fonts */
+body {
+    font-family: 'Roboto', sans-serif;   
+}
+
+
+.home {
```
amellie, line 62: Do we really need a grid here? Each row looks like a normal container/element to me

amellie (commented):
Really great effort on the website! Still a bit of work needs to be done on some sections.

PR #93, opened 2021-12-11
London class 8   ikaneng khoza/responsive grid project

ikanengKhoza:


index.html
```diff
@@ -4,18 +4,185 @@
...
+
+            
+
+        
+       
+    </main>
+
+
+<!--
+<body>
```
amellie: commented codes should always be removed

css/main.css
```diff
@@ -1,3 +1,263 @@
...
+  flex-direction: row;
+  gap: 5px;
+}
+
+.homepage-main {
+  display: grid;
+  grid-template-columns: repeat(2, 1fr);
+  grid-template-rows: 126.4px 952px;
+
+  grid-template-areas:
```
amellie: you define a grid-template-areas but you haven't define grid-area? remove this as it is not used

css/main.css
```diff
@@ -1,3 +1,263 @@
...
+.homepage-content {
+  background-color: var(--secondary);
+  margin-left: 10px;
+  grid-column-start: 1;
+  grid-column-end: -1;
+  padding: 10px;
+  width: 100vw;
+  margin: 20px;
+  width: 280px;
+  height: 375px;
```
amellie: we shouldn't be using height on this, as this is counter-intuitive. As soon as you add a bit of texts, or resize texts, then you need to change the height again.

Removing height will make the section to have a background colour, therefore you need to split the section into two to prevent this from happening (i.e. the Maya Nelson box and the Hello text box)

css/main.css
```diff
@@ -1,3 +1,263 @@
...
+  grid-row: 2;
+}
+
+.homepage-aside {
```
amellie: unused style? remove it if it's not used

css/main.css
```diff
@@ -1,3 +1,263 @@
...
+
+.blue-square {
+  background-color: var(--accent-colour);
+  height: 16px;
+  width: 16px;
+  margin-top: 30px;
+}
+
+.img-hero {
+  width: 120px;
```
amellie: adding width and height directly into an image is not recommended, as I explained previously

index.html
```diff
@@ -4,18 +4,185 @@
...
+            <p class="font-5">Here's who I am and what I do</p>
+            <p>I'm a paragraph. Click here to add your own text and edit me. It’s easy. Just click “Edit Text” or double click me to add your own content and make changes to the font.</p>
+            <p>I’m a great place for you to tell a story and let your users know a little more about you.</p>
+        </div>
+        </section>
+
+       
+       
+       
+        <section class="homepage-footer">
```
amellie: what happens to using `<footer>` tag?

amellie (commented):
Hi Ika, good attempt, but there are a lot of work that still need to be done to this page (as confirmed by you). Will add more comments once you have completed it

PR #94, opened 2021-12-12
LDB 8 Updated  Commit  responsive grid

LovesPictures:


css/style.css
```diff
@@ -0,0 +1,798 @@
...
+    box-sizing: border-box;
+    margin: 0px;
+    padding: 0px;
+}
+
+html {
+    display: block;
+    margin-left: auto;
+    margin-right: auto;
+    max-width: 380px;
```
LucyMac, line 41: You're restricting your page to mobile mode here which is not recommended.

Remember, "mobile first" does not mean limiting to mobile. It means making the mobile design the default for all screen sizes. So you code up your mobile design first, and it looks great on mobile - but since there are no media queries (and no max-width!), it will also work on desktop. It just won't follow the desktop layout until you 'fix' it and add media queries for it to adapt and change on larger screens.

css/style.css
```diff
@@ -0,0 +1,798 @@
...
+    /*font*/
+    color: var(--font-pr-color);
+}
+
+
+h1 {
+    font-family: var(--font-primary);
+    font-size: 2.3em;
+    font-weight: 100px;
+    padding: 1em 0em 0em 0em;
```
LucyMac, line 59: FYI when setting a measurement to `0` you don't need to set the unit. So here you can say `1em 0 0 0`
On another note, using `rem` might be best as a general rule as it will always be based on the base size of the browser (16px by default). 
`em` on the other hand is dependent on the size set on the parent of a given element. So let's imagine this `h1` was inside an element that had `font-size: 24px`, then 1em would equal 24px in this case.

css/style.css
```diff
@@ -0,0 +1,798 @@
...
+
+h6 {
+    font-size: 20px;
+    font-weight: 650;
+    margin: 20px  0px 10px 0px;
+    color: var(--font-pr-color);
+    width: 300px;
+}
+
+h7 {
```
LucyMac, line 98: `h7` doesn't exist I'm afraid :) 

Nice and neat definition of all the other headings though, very organised 👌

css/style.css
```diff
@@ -0,0 +1,798 @@
...
+    font-family: inherit;
+    /* line-height: 1.6em; */
+    margin-block-start: 1em;
+    margin-block-end: 1em;
+    margin-inline-start: 0px;
+    margin-inline-end: 0px;
+    margin-block-start: 1em;
+    margin-block-end: 1em;
+    margin-inline-start: 0px;
+    margin-inline-end: 0px;
```
LucyMac, line 114: There is duplicate code here. Note these particular rules are defaults set by the browser usually. You can stick to just using `margin` and it will override all of these. 
As a general rule I would set `margin` to `0` on generic elements like `p`, and then you can add spacing as needed in different scenarios (unlikely you'll always want 1em of space on either side of all paragraphs).

index.html
```diff
@@ -1,21 +1,156 @@
...
+
+<!--   javasrio fo the returb to the top  -->
+  <link rel="shortcut icon" type="image/x-icon" href="favicon.ico">
 </head>
 
 <body>
+  <main class="grid">
+      
+      <section class="section sh"> 
+        <img src="img/search.png" alt="search" width="18" height="18"><input type="text" id="search" placeholder="Search.." name="search">
```
LucyMac, line 27: Remember to have a `header` before `main`!

LucyMac: Same goes for the footer, I see you don't have one. 
It should go:
`<header></header>`
`<main></main>`
`<footer></footer>`

Inside each of these you can put as many sections as you want.

css/style.css
```diff
@@ -0,0 +1,798 @@
...
+.grid {
+    display: grid;
+    grid-template-columns: repeat(12, minmax(30px, 437px));
+    /* background-color: var(--bg3-highlight-color); */
+}
+
+
+/* Line 1  */
+.sh {
+    grid-area: sh;
```
LucyMac, line 132: None of your `grid-area` rules are working because you haven't set `grif-template-areas` on the parent. 

In any case, you don't need these here since you've opted for `grid-template-columns` (on parent) and `grid-column` (on the children) instead. You don't need both methods.

css/style.css
```diff
@@ -0,0 +1,798 @@
...
+    background-color: var(--bg3-highlight-color);
+}
+
+.nv {
+    grid-area: nv;
+    grid-column: 11/ 13;
+    grid-row: 2;
+    padding: 10px 8px 20px 0px;
+
+    float: none;
```
LucyMac, line 188: You don't need this `float` rule here. You shouldn't ever need to use floats, they are antiquated and have been replaced by flexbox and grid.

index.html
```diff
@@ -1,21 +1,156 @@
...
+             <h1>PRICKLES & CO</h1> 
+                <p>BRING NATURE INDOORS<p> </li> 
+                  <button type="submit" class="shop submit button " onclick="window.open('https://lovespictures.com/','_blank')"> Shop Now</button>
+      </section>
 
-    <!-- Add your markup here -->
 
+      <section class="section i1a "> 
+        <li><h6> C A C T I </h6>  <img src="img/dash.png" alt="logo"  width="30" height="20"></li>
+        <button type="submit" class="shop collection" onclick="window.open('#')"> Shop Collection</button>
```
LucyMac, line 45: For buttons, avoid naming their classes after what they say/do. Most buttons are repeated across a website - perhaps you have 1, 2 or 3 button styles maximum that re-occur on different pages. Because of this, we tend to give buttons class names like `button--primary`, `button--secondary`, or sometimes `button--solid` and `button--outline`... you get the idea. This way they can be re-used throughout, without any concern for what the wording is inside the button.

LucyMac: (in case you're wondering why I'm using `--` in the class names, I'm following the [BEM notation](https://css-tricks.com/bem-101/))

index.html
```diff
@@ -1,21 +1,156 @@
...
+      
+      </section>
+      <section class="section i2a "> 
+        <li><h6> P L A N T S </h6>  <img src="img/dash.png" alt="logo"  width="30" height="20"></li>
```
LucyMac, line 49: You can achieve the spaced out letters but using the `letter-spacing` CSS rule. Better than doing it in the HTML as it is just a stylistic feature.

css/style.css
```diff
@@ -0,0 +1,798 @@
+@import url('https://fonts.googleapis.com/css2?family=Roboto:wght@300&display=swap');
```
LucyMac, line 1: I see you're importing your fonts in 2 different ways: with @import here and also in your HTML `<head>`. You only need one of these (probably stick to `head` for now)

The fonts for this project should be different as described in the readme - can you change them? It will make a world of difference to how your site looks 🙂

LucyMac (commented):
Nice work using grid and CSS variables, and generally having clean and organised code. 

Watch out with how you've restricted the screen width - your site should work visually for desktop too and fill the screen, see my comments below. 

Keep going!

PR #95, opened 2021-12-12
London Class 8 - Ali Jahanakah -  responsive-grid-project - Week 4

Ali-Jahankah:


css/common.css
```diff
@@ -0,0 +1,52 @@
+:root {
+  --gray: #ededee;
+  --red: #a05941;
+  --black: #133032;
+  --white: #fff;
+  --transwhite: rgba(0, 0, 0, 0.75);
```
LucyMac, line 6: 👍

css/common.css
```diff
@@ -0,0 +1,52 @@
...
+  padding: 0;
+  margin: 0;
+  border: 0;
+  box-sizing: border-box;
+  outline: 0;
+}
+body {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
```
LucyMac, line 18: Be careful setting rules that are a little too specific on the body. 
Bear in mind this will apply only to header, main and footer levels, nothing lower (more nested) than that. These will be stacked by default (`display: block` is the default), and this won't change across devices - so are these rules needed?

css/common.css
```diff
@@ -0,0 +1,52 @@
...
+  background: var(--red);
+  color: var(--white);
+}
+h1 {
+  font-size: 6vw;
+  font-family: "Fjalla One", sans-serif;
+}
+h2 {
+  font-size: 4.5vw;
+  font-family: "Outfit", sans-serif;
```
LucyMac, line 33: The header h2 should be Fjalla font. Add a class to it so you can change it without affecting other `h2`s

css/common.css
```diff
@@ -0,0 +1,52 @@
...
+    font-size: 2.5vw;
+  }
+  h2 {
+    font-size: 2vw;
+  }
+  h3 {
+    font-size: 1vw;
+  }
+  .btn {
+    font-size: 2vw;
```
LucyMac, line 50: Although having responsive fonts is a nice concept, in practice this does not remove the need to change these at different breakpoints. They tend to get too small on small screens, and too large on large screens. I would argue you have more control over this if you set fixed font-sizes (using `rem`) and changing them at different breakpoints (at least for the smallest size and largest screen size).

css/main.css
```diff
@@ -1,3 +1,245 @@
...
+.aboutus {
+  width: 100%;
+  display: grid;
+  grid-template-columns: repeat(4, 1fr);
+  grid-template-rows: repeat(7, 1fr);
+  text-align: center;
+  color: var(--red);
+  line-height: 1.5em;
+  font-size: 5vw;
+  font-weight: bold;
```
LucyMac, line 104: Nice grid! Looks great 👌

LucyMac: (color: red??) 😳

css/footer.css
```diff
@@ -0,0 +1,140 @@
...
+.footer_item ul li {
+  list-style: none;
+  margin-top: 5px;
+}
+.footer_item ul li a {
+  color: white;
+}
+.footer_item h1 {
+  font-size: 8vw;
+}
```
LucyMac, line 35: The font should be different here. 
Also, I wouldn't use an `h1` here - keep `h1` for the main large title on the page, and use `h2`-`h6` for the rest. A good way to decide which heading tag to use is to make a sort of hierarchy of all the headings in a design, and order their styles by 'importance' (size, thickness etc). Perhaps this is an h4 for example, if compared to the other styles on the page.

css/footer.css
```diff
@@ -0,0 +1,140 @@
...
+.social_icons ul li {
+  list-style: none;
+}
+.social_icons ul li a {
+  display: block;
+}
+.social_icons ul li a img {
+  width: 50%;
+  max-width: 100%;
+}
```
LucyMac, line 93: It's generally considered best practice to avoid too much nested selectors. If you want to target this kind of element, it might be best just to give the `img` its own class, and target that directly.

css/main.css
```diff
@@ -1,3 +1,245 @@
...
+  .third_hero_section h2 {
+    font-size: 5vw;
+  }
+  .cacti,
+  .plants,
+  .succulents {
+    width: 30%;
+    padding: 1em 0;
+    background-position: center;
+  }
```
LucyMac, line 200: Why not target the `product_item` class directly here? 

As a suggestion, using grid might be easier here. 3 columns of `1fr` each, a column-gap, and you're done 😉

css/common.css
```diff
@@ -0,0 +1,52 @@
...
+}
+h2 {
+  font-size: 4.5vw;
+  font-family: "Outfit", sans-serif;
+}
+input {
+  padding: 0.5em 1em;
+}
+/* ============================Media Queries========================= */
+@media screen and (min-width: 481px) {
```
LucyMac, line 39: 481px is a little too early to be switching to desktop layout. I'd keep the mobile layout until 600px minimum. Then change a few things for tablet (even if jsut a few things like font-sizes), and go full desktop above 1000px.

LucyMac (commented):
Nice work @Ali-Jahankah. A good effort to structure and style your page appropriately. There are a few inconsistencies with the design, mostly font families and sizes, which could be tidied up - see my comments. 
I would recommend increasing your 481px breakpoint a little, and adding an extra one as well. This will enable you to better control things like icon and font sizes which aren't working at all screen sizes currently. 
Can you add another page as well? 😉

css/common.css
Ali-Jahankah: Hi Lucy, Thank you for your feedback. I'll do that 🙌

css/common.css
Ali-Jahankah: Yes you right Lucy. I'll change that as well. Thank you!

css/common.css
Ali-Jahankah: Sorry I forgot to delete that because I used to give min-height to Main then I removed that and did this to avoid any mess up later. I'll remove that . Thanks!

css/main.css
Ali-Jahankah: Yes you're right 😃 Thanks!

PR #96, opened 2021-12-13
London-Class-8/Rahwa-Ghebremichael/Responsive-grid-project

rahwafesseha:
I did the mobile version only.


css/about.css
```diff
@@ -0,0 +1,61 @@
...
+  box-sizing: border-box;
+  margin:0;
+  padding:0;
+  
+}
+​
+body{
+  
+   font-family: 'Quicksand', sans-serif; 
+  background: var(--light-beige);
```
mahsa2: Well done on using variables 🌟

index.html
```diff
@@ -5,17 +5,144 @@
...
+              <li>Resume</li>
+              <li>Projects</li>
+              <li>Contacts</li>
+            </ul>
+          </nav>
+        </section>
+      </header>
+​
+     <main>
+        <div class="my-image">
```
mahsa2, line 50: To put an absolute image in the center, you can have a look at this answer: https://stackoverflow.com/a/8273750/1189105

mahsa2 (commented):
Well done on working on your assignment.

I made a commit on your project to save what we talked about on your project and how you should approach certain situations. I would recommend to fix the points we talked today to get a better understanding of what we talked about and learn in practice if you can.

The key points are:
- Use common.css to reuse the common use between pages.
- Make sure the html contents are properly closed, and have the right indentations.
- Use grid whenever you need to fit an item
- Make sure your website is identical to the design.

PR #139, opened 2022-07-29
design_2_resume_franklin

FranklinPTorres:


index.html
```diff
@@ -4,18 +4,165 @@
...
+                </ul>
+            </nav>
+        </header>
+
+        
+
+    <!-- MAIN -->
+<div class="mobile">
+
+    <div class="section_2">
```
jcholyhead, line 58: if it's called "section", maybe a section tag would be more semantic

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
```
jcholyhead, line 3: add a .gitignore file

css/second.css
```diff
@@ -0,0 +1,359 @@
...
+    }
+
+        .content-2 {
+            display: flex;
+            justify-content: center;
+            gap: 450px;
+    
+        }
+
+        .desktop-educ{
```
jcholyhead, line 335: what's going on with the indentation in this file?

jcholyhead (commented):
Good. The site is responsive, and looks true to the original design.

PR #169, opened 2022-08-18
Nw5_Leeds_Meron_Gebremichael_html/css_week4

Meronggk:


levilu800b (commented):
Hello Meron, you have done a good job with building up your project. I think you are missing out the responsiveness. I would suggest mobile first when working on frontend for net time.

PR #178, opened 2022-08-19
NW-5 Manchester KHADAR-M-DAGAL- HTML&CSS W4 (Responsive-Grid-Project)

khmdagal:


Ara225:
In terms of actual code not much to fault here, and the commit messages are impecable. You're starting to find a really good balence in comments as well, explaining the strange things while not explaining the obvious, which I personally like. It's fully responsive as well which is wonderful! The design does need polish to fully match the orginal, which was the spec, the orginal has a lot of tiny nice touches like the box shadows on the resume, the blue highlights, a sans serif font and a lot more whitespace

I wanted to mention a few nitpicky things I think you should change to make the code a tinsy bit better, but these are VERY small non functional complaints. It is overall a solid piece of work from a code point of veiw. You should be very proud!
- Good that you're putting alt text in for images but it'd be better to put it the way you would in a sentence so it makes more sense for a screenreader e.g. Facebook logo rather than facebookImage
- Did notice a few spelling mistakes (really not that many at all tho), so I wanted to metion a vscode spell checking plugin that really helps me (my spelling is REALLY bad) https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker 
- Format of class names is inconsistent (sometimes camelcase e.g. profilePhoto, sometimes snakecase e.g.  profile_contianer sometimes a mix e.g. blueBox_Name_and_JobTitle), it's nice to pick one format and stick with it if possible, it just looks less sloppy

PR #181, opened 2022-08-19
north-west-5/shimenafshar/week4

ShimenAfshar:


css/resume.css
```diff
@@ -0,0 +1,122 @@
...
+}
+
+nav a {
+  color: black;
+  text-decoration: none;
+}
+
+/* Hamburger Menu Start */
+
+.ham_menu {
```
Nomes27, line 64: nice use of a responsive hamburger menu :)

index.html
```diff
@@ -1,21 +1,148 @@
...
+          <div class="line-menu half start"></div>
+          <div class="line-menu"></div>
+          <div class="line-menu half end"></div>
+        </div>
+      </div>
 
+      <p>PROJECT MANAGER</p>
+      <ul id="menu" class="menu">
+        <li><a href="index.html">ABOUT ME</a></li>
+        <li><a href="./css/resume.html">RESUME</a></li>
```
Nomes27, line 40: your link looks to be pointing to the wrong location, this should just be resume.html

index.html
```diff
@@ -1,21 +1,148 @@
...
+        <li><a href="#">PROJECTS</a></li>
+        <li><a href="#">CONTACT</a></li>
+      </ul>
+    </nav>
+  </header>
+
+  <!------main----->
+  <main>
```
Nomes27, line 48: nice job using headers & main to split your html content

index.html
```diff
@@ -1,21 +1,148 @@
...
+    <h4>Here's who I am & what I do</h4>
+    <p> I'm a paragraph. Click here to add your own text and edit me. Its easy.
+      Just click “Edit Text” or double click me to add your own content and
+      make changes to the font. Im a great place for you to tell a story and
+      let your users know a little more about you.</p>
+  </section>
+</main>
+<!-----footer----->
+  <footer>
+    <div class="call_div center">
```
Nomes27, line 106: nice creation of a reusable `center` class

resume.html
```diff
@@ -0,0 +1,218 @@
+<!DOCTYPE html>
```
Nomes27, line 1: your resume page looks great on mobile, but looks to be a bit broken on desktop size

resume.html
```diff
@@ -0,0 +1,218 @@
...
+        <h2>Experience</h2>
+        <a
+          href="https://3f62d478-a066-43ee-98e6-3e48600ff893.filesusr.com/ugd/84770f_5db758741cef46fbb32e0125937f8de5.pdf"
+          target="_blank"
+        >
+          <button class="dwnload_btn">DOWNLOAD CV</button></a
+        >
+      </section>
+      <!-- ARTICLE START -->
+      <article class="art1">
```
Nomes27, line 54: instead of using classes like `art1`, `art2`, `art3` etc I would try to create just one reusable class like you have done with `art`, the articles look like they are using similar styles so you could just use one class name for these

index.html
```diff
@@ -1,21 +1,148 @@
...
+            src="https://static.wixstatic.com/media/01c3aff52f2a4dffa526d7a9843d46ea.png/v1/fill/w_20,h_20,al_c,q_85,usm_0.66_1.00_0.01/01c3aff52f2a4dffa526d7a9843d46ea.webp" />
+        </a>
+      </div>
+    </div>
+    <div class="content">
+      <h1>Hello</h1>
+      <h4>Here's who I am & what I do</h4>
+      <div class="button_div">
+        <button>RESUME <a href="/resume.html" class="card-btn_1"></a></button>
+        <button>PROJECTS <a href="/projects.html" class="card-btn_2"></a></button>
```
Nomes27, line 86: you should just use an `<a>` tag here rather than an `a` tag inside a button, `a` tags are used for navigation so this would be the best semantic html to use

Nomes27:
Great job Shimen, you've followed the given design really closely and you've used responsive css really well for your index page :) there just looks to be one issue with your resume page for desktop size, but you've made a great start with the mobile design

PR #185, opened 2022-08-19
NW5-Manchester-ANVAR AZIZI-HTML&CSS-Week 4

rebwar1:


KarenPudner (approved):
This looks professional - there is a small issue with text "...a little more about you" not quite fitting the beige background when resized to the smallest possible window, but other than that the resizing works well. I also like how the navigation links appear when the "hamburger" is clicked. I can't see any use of CSS grid, as specified in the instructions - was it a deliberate choice not to use this?

rebwar1:
> This looks professional - there is a small issue with text "...a little more about you" not quite fitting the beige background when resized to the smallest possible window, but other than that the resizing works well. I also like how the navigation links appear when the "hamburger" is clicked. I can't see any use of CSS grid, as specified in the instructions - was it a deliberate choice not to use this?

rebwar1:
> This looks professional - there is a small issue with text "...a little more about you" not quite fitting the beige background when resized to the smallest possible window, but other than that the resizing works well. I also like how the navigation links appear when the "hamburger" is clicked. I can't see any use of CSS grid, as specified in the instructions - was it a deliberate choice not to use this?

I thought it's not complicated so because of that I didn't use grid for lay out but still it's not finish I'll work on it till complete the project and also I should working on sass and refactor it.

PR #197, opened 2022-09-30
ZA2-Patou-Mabando-responsive-grid-project

PatouMabando:


contact.html
```diff
@@ -0,0 +1,116 @@
...
+   <h1>Let's talk</h1>
+  </div>
+  <div class="fieldset__container">
+   <fieldset>
+    <div>
+     <div class="form_column">
+      <label for="firstname">First name*</label>
+      <input id="firstname" type="text" required size="12" name="firstname" /><br>
+     </div>
+     <div class="form_column">
```
pkijowska, line 65: I noticed there is only one _ and was wondering if it is intentional. If you want to use BEM structure then __ would be prefered like you did with fieldset__container.

css/contact.css
```diff
@@ -0,0 +1,321 @@
...
+ margin-left: -2em;
+ margin-top: -45px;
+ width: 2em;
+ height: 45px;
+ z-index: 5;
+}
+
+.hamburger div {
+ position: relative;
+ width: 3em;
```
pkijowska, line 63: in my opinion use either em or px, I noticed you used both

css/contact.css
```diff
@@ -0,0 +1,321 @@
...
+ grid-template-rows: repeat(4, auto);
+ grid-template-areas:
+  "header header header header"
+  ". text_top text_top ."
+  ". form form ."
+  "footer footer footer footer"
+ ;
+}
+
+/* header */
```
pkijowska, line 201: have you tried to install prettier or any other file formatter as adding an indent would be good in here

css/header.css
```diff
@@ -0,0 +1,129 @@
...
+.blue__box {
+ margin-right: 0.5rem;
+ margin-top: 0.35rem;
+ height: 0.9rem;
+ width: 0.9rem;
+ background-color: var(--third-color-light-beige);
+}
+
+.job__title_container {
+ grid-area: j;
```
pkijowska, line 58: I think job__title_container should be job__title__container or job__title-container

pkijowska (commented):
I really like what you did. The only thing is that I am unable to see the image but overall, the project looks amazing. Could be a great portfolio piece. Make sure you remember to indent and think about using correct BEM structure.

pkijowska (approved):

PR #202, opened 2022-11-16
London 9 Turing - susan - HTML/ CSS - Week 4

susanssky:
https://hkuhku609.github.io/responsive-grid-project/


index.html
```diff
@@ -1,21 +1,214 @@
...
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Responsive grid project</title>
-</head>
+    <link rel="stylesheet" href="./css/main.css" />
+    <link
+      rel="stylesheet"
+      href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.2.0/css/all.min.css"
```
SallyMcGrath: My guess is that if you ran Lighthouse you would get a warning telling you not to chain your critical requests. 

CSS is "render blocking", so the browser will stop doing everything else until it's done with the CSS. Have a look on webpagetest to learn a bit more about this https://www.webpagetest.org/

index.html
```diff
@@ -1,21 +1,214 @@
...
+
+  <body>
+    <header>
+      <div class="bottom-btn">
+        <i class="fa-solid fa-comment"></i>
+        <p class="desktop-letschat">Let's Chat</p>
+      </div>
+      <section class="header-section">
+        <div class="header-search">
+          <input type="text" placeholder="Search..." />
```
SallyMcGrath: This input isn't labelled, but we know that all inputs must be labelled. It's a tricky challenge! There's a useful pattern for labelling things without showing the text. Look up https://webaim.org/techniques/css/invisiblecontent/

index.html
```diff
@@ -1,21 +1,214 @@
...
 
-<body>
+          <i class="fa-solid fa-bag-shopping">
+            <div class="cart-num">0</div>
+          </i>
+        </div>
+      </section>
+      <nav>
+        <div class="nav-title">
+          <p class="fjalla">prickles & co.</p>
```
SallyMcGrath: Is there a semantic html element for a title? What is another word for title?

The nav element will show that this is a nav block, but it won't show what the subject of the navigation is. 

https://www.w3.org/WAI/ARIA/apg/example-index/landmarks/HTML5.html

index.html
```diff
@@ -1,21 +1,214 @@
...
+          <ul class="desktop-nav">
+            <li>Shop</li>
+            <!-- <li>Cacti</li>
```
SallyMcGrath: Whoops! looks like some commented out code has been left in here.

index.html
```diff
@@ -1,21 +1,214 @@
...
+            <li>Plants</li>
+            <li>Succulents</li> -->
+            <li>About</li>
+            <li>FAQ</li>
+            <li>Contact</li>
+          </ul>
+        </div>
 
-    <!-- Add your markup here -->
+        <div class="nav-tel">
```
SallyMcGrath: Could this be a tel link? How do those work?

index.html
```diff
@@ -1,21 +1,214 @@
...
+        <div class="main-collection-box">
+          <h2>succulents</h2>
+          <h2>-</h2>
+          <button>Shop Collection</button>
+        </div>
+      </section>
+      <section class="main-about-shop">
+        <div class="main-about">
+          <p class="main-spacing">about us</p>
+          <h2>-</h2>
```
SallyMcGrath: In a headings-only view of this page, what will this -- communicate to the user?

https://usability.yale.edu/web-accessibility/articles/headings

index.html
```diff
@@ -1,21 +1,214 @@
...
+          </p>
+          <p>
+            This is a great space to write long text about your company and your
+            services. You can use this space to go into a little more detail
+            about your company. Talk about your team and what services you
+            provide.
+          </p>
+        </div>
+        <div class="mobile-main-pics">
+          <img src="./img/mobile/4.webp" alt="4" /><img
```
SallyMcGrath: Is this alt text useful? How will it communicate the subject of the picture to a user who cannot access the image?

index.html
```diff
@@ -1,21 +1,214 @@
...
+        <h1>-</h1>
+        <form action="">
+          <div class="fn">
+            <label for="">First Name *</label> <input type="text" />
+          </div>
+          <div class="ln">
+            <label for="">Last Name *</label> <input type="text" />
+          </div>
+          <div class="em">
+            <label for="">Email *</label> <input type="text" />
```
SallyMcGrath: Form controls must be associated properly with labels. 

https://dequeuniversity.com/rules/axe/4.4/label

Why does this matter? Can you take a guess? (OK to not know!)

index.html
```diff
@@ -1,21 +1,214 @@
...
+          <button class="bu">Send</button>
+        </form>
+      </section>
+      <section class="main-bg"></section>
+    </main>
+    <footer>
+      <section class="footer-section">
+        <div class="footer-3details">
+          <div class="footer-detail">
+            <p>OUR STORE</p>
```
SallyMcGrath: 1. Is there a (or 6) semantic html element that could go here instead?
2. Is there a different way to uppercase headings? What is the impact of hard coding text in all caps vs using CSS to style it?

css/main.css
```diff
@@ -1,3 +1,555 @@
...
+.main-collection-box h2 {
+  font-family: var(--fjalla-one);
+  font-size: 1.2rem;
+  letter-spacing: 0.1rem;
+  color: var(--red-brown);
+}
+.desktop-letschat {
+  font-family: var(--fjalla-one);
+}
+.desktop-letschat,
```
SallyMcGrath, line 43: I wonder, is there a helper or "utility class" you could use here to hide things on mobile and apply it to the elements directly? Have a google and think if there's something else you might find useful to do here.

css/main.css
```diff
@@ -1,3 +1,555 @@
...
+  color: var(--red-brown);
+  text-align: center;
+  font-size: 2rem;
+}
+.main-order form {
+  display: grid;
+  padding: 1rem 1rem;
+  width: 100%;
+  gap: 1rem;
+  grid-template-areas:
```
SallyMcGrath, line 352: We write code for other people to read. Please use real words to describe your grid areas, so you can get the benefit of named areas, which is the self-documenting nature of the spec. 

Use whole words always in named areas. There are tools like cssnano to compress your code, after you have written it, in production, but you shouldn't try to do this manually.

css/main.css
```diff
@@ -1,3 +1,555 @@
...
+  .desktop-pic4 {
+    display: block;
+    grid-area: p4;
+    background-image: url('../img/desktop/4.webp');
+    background-size: cover;
+  }
+  .desktop-pic5 {
+    display: block;
+    grid-area: p5;
+    grid-column: 4/5;
```
SallyMcGrath: Typically, you will either use grid-area spec or assign to rows and columns. You shouldn't generally need to do both. Can you talk a little about why you chose to do this here?

SallyMcGrath (commented):
Fab work! Super impressive. I've dropped a few quick notes from reading through, and it was super easy to eyeball the look as you deployed to github pages. 

Really ace. Thank you!

Please commit much more frequently and describe your commits clearly. 10% done is not very illuminating! https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/

index.html
```diff
@@ -82,63 +117,95 @@ <h2>-</h2>
...
-          src="./img/mobile/5.webp"
-          alt=""
-        /><img src="./img/mobile/6.webp" alt="" />
-        <h1>BUY ONLINE NOW & GET 10% OFF !</h1>
-        <button>Shop Now</button>
+        <div class="mobile-main-pics">
+          <img src="./img/mobile/4.webp" alt="4" /><img
+            src="./img/mobile/5.webp"
+            alt="5"
+          /><img src="./img/mobile/6.webp" alt="6" />
```
samhoooo: You should specify explicit width and height on   to avoid cumulative content shift. Reference: https://www.aleksandrhovhannisyan.com/blog/setting-width-and-height-on-images/

index.html
susanssky: Hi. Do you refer "h1"?

SallyMcGrath: https://www.w3.org/WAI/tutorials/page-structure/headings/

Depending on what is already on the page, I might use an h2 or h3. It's definitely operating as a heading/title in this block, though, and not a paragraph.

index.html
susanssky: for screen readers to read. :)

css/main.css
susanssky: Sorry, personal habit, it has been removed now

susanssky:
completed

PR #210, opened 2022-11-18
 London9-Leila Farsani-responsive-grid-project-Week 4

leilafarsani:
Deployed Link:
https://stellar-squirrel-248391.netlify.app/

### DIFF
--- a/index.html
+++ b/index.html
@@ -85,6 +85,31 @@
         </div>
       </div>
     </main>
-    <!-- Please leave this in to credit the creators of this design -->
+    <footer>
+      <div class="footer-call">
+        <div class="footer-major">Call</div>
+        <a class="footer-contain" href="tel:+441234567890">123-456-7890</a>
+      </div>
+      <div class="footer-write">
+        <div class="footer-major">Write</div>
+        <a class="footer-contain" href="mailto:info@mysite.com"
+          >info@mysite.com</a
+        >
+      </div>
+      <div class="footer-follow">
+        <div class="footer-major">Follow</div>
+        <div class="footer-contact">
+          <i class="fa-brands fa-facebook-f"></i>
+          <i class="fa-brands fa-twitter"></i>
+          <i class="fa-brands fa-linkedin-in"></i>
+          <i class="fa-brands fa-instagram"></i>
+        </div>
+      </div>
+      <!-- Please leave this in to credit the creators of this design -->
+      <p class="purpose">
+        This website design was created by Wix.com, and is used here for
+        strictly educational purposes.
+      </p>
+    </footer>
   </body>
 </html>
--- a/index.html
+++ b/index.html
@@ -5,18 +5,46 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <!-- Add a link to your custom CSS here -->
-    <link rel="stylesheet" href="homepage.css" />
+    <link rel="stylesheet" href="style.css" />
     <title>Resume</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Poppins:wght@600&display=swap"
+      rel="stylesheet"
+    />
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Quicksand:wght@300&display=swap"
+      rel="stylesheet"
+    />
   </head>
-  
-
   <body>
-    <!-- Add your markup here -->
-
+    <header>
+      <div class="header-box">
+        <div class="box1-name">
+          <div class="box1"></div>
+          <div class="header-name">
+            Maya Nelson
+            <span class="header-title"> / PROJECT MANAGER</span>
+          </div>
+        </div>
+        <div class="header-navigation">
+          <ul class="header-nav-menu">
+            <li><a href=""></a>ABOUT ME</li>
+            <li><a href=""></a>RESUME</li>
+            <li><a href=""></a>PROJECTS</li>
+            <li><a href=""></a>CONTACT</li>
+          </ul>
+        </div>
+        <div class="hamburger-nav-menu">
+          <hr class="line1" />
+          <hr class="line2" />
+          <hr class="line3" />
+        </div>
+      </div>
+    </header>
     <!-- Please leave this in to credit the creators of this design -->
-    <p>
-      This website design was created by Wix.com, and is used here for strictly
-      educational purposes.
-    </p>
   </body>
 </html>
--- a/index.html
+++ b/index.html
@@ -86,9 +86,9 @@
         <div class="hello-part">
           <div class="hello-text">Hello</div>
           <div class="role">Here's who I am & what I do</div>
-          <div class="btns">
-            <button class="btn" class="resume-btn">RESUME</button>
-            <button class="btn" class="projects-btn">PROJECTS</button>
+          <div class="btn">
+            <button class="resume-btn">RESUME</button>
+            <button class="projects-btn">PROJECTS</button>
           </div>
           <p class="p1">
             I'm a paragraph. Click here to add your own text and edit me. It’s
--- a/index.html
+++ b/index.html
@@ -32,12 +32,13 @@
         </div>
         <div class="header-navigation">
           <ul class="header-nav-menu">
-            <li><a href=""></a>ABOUT ME</li>
+            <li class="about-me"><a href=""></a>ABOUT ME</li>
             <li><a href=""></a>RESUME</li>
             <li><a href=""></a>PROJECTS</li>
             <li><a href=""></a>CONTACT</li>
           </ul>
         </div>
+
         <div class="hamburger-nav-menu">
           <hr class="header-line1" />
           <hr class="header-line2" />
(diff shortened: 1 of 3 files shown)


leilafarsani:
https://stellar-squirrel-248391.netlify.app/

index.html
```diff
@@ -1,21 +1,152 @@
...
+      href="https://fonts.googleapis.com/css2?family=Quicksand:wght@300&display=swap"
+      rel="stylesheet"
+    />
+  </head>
+  <body>
+    <header>
+      <div class="header-box">
+        <div class="box1-name">
+          <div class="box1"></div>
+          <div class="header-name">
```
SallyMcGrath, line 28: Is there a semantic html element you might use here to signify this heading?

index.html
```diff
@@ -1,21 +1,152 @@
...
+        </div>
+      </div>
+    </header>
+    <main>
+      <div class="background1"></div>
+      <div class="main-content">
+        <div class="main-box-container">
+          <img
+            src="https://static.wixstatic.com/media/2e2a49_bd3a1599bc1b4265a01b2754b99b8536~mv2_d_5760_3840_s_4_2.jpeg/v1/crop/x_2703,y_43,w_2199,h_2205/fill/w_205,h_205,al_c,q_80,usm_0.66_1.00_0.01,enc_auto/AdobeStock_230407429.jpeg"
+            alt="Maya Photo"
```
SallyMcGrath, line 55: Nice alt text! To polish this even further, what about saying "A photo of Maya." or "A photo of Maya, a young white woman with brown hair cut in a cute bob"

index.html
```diff
@@ -1,21 +1,152 @@
...
+              src="https://static.wixstatic.com/media/01c3aff52f2a4dffa526d7a9843d46ea.png/v1/fill/w_25,h_25,al_c,q_85,usm_0.66_1.00_0.01,enc_auto/01c3aff52f2a4dffa526d7a9843d46ea.png"
+              style="width: 25px; height: 25px; object-fit: cover"
+            />
+          </div>
+        </div>
+        <div class="hello-part">
+          <div class="hello-text">Hello</div>
+          <div class="role">Here's who I am & what I do</div>
+          <div class="btn">
+            <button class="resume-btn">RESUME</button>
```
SallyMcGrath, line 90: Should this be a button or a link _styled_ as a button?

Buttons trigger actions; links navigate to new locations. What would you choose?

index.html
```diff
@@ -1,21 +1,152 @@
...
+      </div>
+    </main>
+    <footer>
+      <p class="purpose1">
+        This website design was created by Wix.com, and is used here for
+        strictly educational purposes.
+      </p>
+      <div class="footer-call">
+        <div class="footer-major">Call</div>
+        <a class="footer-contain" href="tel:+441234567890">123-456-7890</a>
```
SallyMcGrath, line 112: Nice use of tel attribute!

style.css
```diff
@@ -0,0 +1,281 @@
...
+  font-weight: bold;
+}
+
+.btn {
+  display: flex;
+  gap: 1rem;
+}
+
+.resume-btn {
+  border-radius: 20px;
```
SallyMcGrath, line 133: You might be fighting here a little against browser styles. This might be interesting to dig into even if you choose not to use the button element. 

Look into 
https://caniuse.com/css-appearance
https://css-tricks.com/overriding-default-button-styles/

style.css
```diff
@@ -0,0 +1,281 @@
...
+
+.footer-follow {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-around;
+  align-items: center;
+}
+
+
+.footer-write {
```
SallyMcGrath, line 179: Is there a way you could combine these two rulesets to make your styles simpler and clearer?

style.css
```diff
@@ -0,0 +1,281 @@
...
+  padding-top: 1rem;
+  gap: 2.5rem;
+}
+img {
+  width: 50%;
+  height: auto;
+  border-radius: 50%;
+}
+.main-name {
+  font-size: 1.6rem;
```
SallyMcGrath, line 81: Look into line-height and then look again at the reference layout. 

https://developer.mozilla.org/en-US/docs/Web/CSS/line-height

What might you set here to get closer to the original image?

style.css
```diff
@@ -0,0 +1,281 @@
...
+
+hr.blue-line {
+  background-color: #0150fd;
+  width: 19%;
+  height: 2px;
+}
+
+.main-title {
+  font-size: 1.1rem;
+  letter-spacing: 0.5rem;
```
SallyMcGrath, line 94: This letter spacing seems very large and the text is hitting the container. What about reducing the spacing and putting some padding on this box somewhere?

index.html
```diff
@@ -1,21 +1,152 @@
...
+      <div class="footer-write">
+        <div class="footer-major">Write</div>
+        <a class="footer-contain" href="mailto:info@mysite.com"
+          >info@mysite.com</a
+        >
+      </div>
+      <div class="footer-follow">
+        <div class="footer-major">Follow</div>
+        <div class="footer-contact">
+          <img
```
SallyMcGrath, line 123: What might these images be meant to do? Is there another element that might offer more functions?

style.css
```diff
@@ -0,0 +1,281 @@
...
+  font-family: "Quicksand", sans-serif;
+}
+
+.contact {
+  background-color: white;
+  align-items: center;
+  width: 100%;
+  display: flex;
+  justify-content: center;
+  height: 3rem;
```
SallyMcGrath, line 104: Typically we avoid setting explicit heights in layout. Could you think about using gap and padding to create this component instead?

SallyMcGrath (commented):
Lovely work, @leilafarsani, and good effort at getting close to the design. Excellent scores on Lighthouse on your deployed page! 🎉 

I see that the Wix site uses Avenir, which is a pro font. I might opt for something close like Montserrat or Manrope, on Google fonts, to replace it, rather than leaving the main body as Times/ default. Take a look and see what you think.

https://fonts.google.com/specimen/Manrope
https://fonts.google.com/specimen/Montserrat

index.html
```diff
@@ -1,21 +1,152 @@
...
     <!-- Add a link to your custom CSS here -->
+    <link rel="stylesheet" href="style.css" />
+    <title>Resume</title>
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
+    <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin />
+    <link
+      href="https://fonts.googleapis.com/css2?family=Poppins:wght@600&display=swap"
+      rel="stylesheet"
+    />
+    <link rel="preconnect" href="https://fonts.googleapis.com" />
```
jonnywyatt, line 16: These two `preconnect` link tags are a good addition and will make the font files load more quickly. 
You only need them once in the `head`, above the first font stylesheet

index.html
```diff
@@ -1,21 +1,152 @@
...
+    <header>
+      <div class="header-box">
+        <div class="box1-name">
+          <div class="box1"></div>
+          <div class="header-name">
+            Maya Nelson
+            <span class="header-title"> / PROJECT MANAGER</span>
+          </div>
+        </div>
+        <div class="header-navigation">
```
jonnywyatt, line 33: This element contains navigation, there's a semantic tag you can use instead of `div`

index.html
```diff
@@ -1,21 +1,152 @@
...
+          <ul class="header-nav-menu">
+            <li class="about-me"><a href=""></a>ABOUT ME</li>
```
jonnywyatt, line 35: These `a` tags should wrap the text, at the moment they're empty

index.html
```diff
@@ -1,21 +1,152 @@
...
-</html>
\ No newline at end of file
+        <div class="hamburger-nav-menu">
+          <hr class="header-line1" />
+          <hr class="header-line2" />
+          <hr class="header-line3" />
+        </div>
+      </div>
+    </header>
+    <main>
```
jonnywyatt, line 49: Nice 👏

index.html
```diff
@@ -1,21 +1,152 @@
...
+              style="width: 25px; height: 25px; object-fit: cover"
+            />
+            <img
+              alt="Instagram"
+              src="https://static.wixstatic.com/media/01c3aff52f2a4dffa526d7a9843d46ea.png/v1/fill/w_25,h_25,al_c,q_85,usm_0.66_1.00_0.01,enc_auto/01c3aff52f2a4dffa526d7a9843d46ea.png"
+              style="width: 25px; height: 25px; object-fit: cover"
+            />
+          </div>
+        </div>
+        <div class="hello-part">
```
jonnywyatt, line 86: Every page should have an `h1`. It should be the text that best describes this page, as distinct from the others in the site. So if all the pages are about Maya Nelson, then the text "Maya Nelson" won't be the best choice. Which heading on this page best describes what this page is about? That should be the `h1`.

index.html
```diff
@@ -1,21 +1,152 @@
...
+          >info@mysite.com</a
+        >
+      </div>
+      <div class="footer-follow">
+        <div class="footer-major">Follow</div>
+        <div class="footer-contact">
+          <img
+            alt="Facebook"
+            src="https://static.wixstatic.com/media/0fdef751204647a3bbd7eaa2827ed4f9.png/v1/fill/w_19,h_19,al_c,q_85,usm_0.66_1.00_0.01,enc_auto/0fdef751204647a3bbd7eaa2827ed4f9.png"
+            style="width: 19px; height: 19px; object-fit: cover"
```
jonnywyatt, line 126: Try to avoid inline style attributes as it means you're repeating the same style for each of the icons. If you use a classname and put the styles in the stylesheet, there's less duplication.

style.css
```diff
@@ -0,0 +1,281 @@
...
+  gap: 1rem;
+}
+
+.purpose2{
+  display: none;
+}
+.hamburger-nav-menu {
+  display: none;
+}
+@media screen and (max-width: 600px){
```
jonnywyatt, line 204: To do 'mobile-first' styling, all the default styles (that aren't in a media query) will get used for small screens. Then each of your media queries should use `min-width` rather than `max-width`, so they are used for larger screens.

style.css
```diff
@@ -0,0 +1,281 @@
...
+  .header-box{
```
jonnywyatt, line 205: This hides the nav (About me, etc) for small screens. Your mobile users will want to be able to navigate between the pages on the site, so better to show the nav for all screen sizes?

style.css
```diff
@@ -0,0 +1,281 @@
...
+      display: none;
+  }
+
+  .hamburger-nav-menu {
```
jonnywyatt, line 209: This isn't having any effect because it's inside `.header-box` which was hidden.

style.css
```diff
@@ -0,0 +1,281 @@
...
+    display: block;
+    margin-right: 3rem;
+  }
+  
+  .header-line1 {
```
jonnywyatt, line 214: The 3 `.header-line` classes aren't being used because they're inside a hidden element (`.header-box`). 
They share several properties eg `position: relative`, `height: 4px` etc. You could extract those shared properties to a class and add that to each of the elements, the way we did with the buttons on our call the other day?

style.css
```diff
@@ -0,0 +1,281 @@
...
+}
+.background1 {
+  background: #e6dacd;
+  width: 40%;
+  height: 90vh;
+  position: absolute;
+  z-index: -1;
+}
+.main-content {
+  display: flex;
```
jonnywyatt, line 60: Remember from the last PR review, for elements to stack vertically, you don't need to use flex, if they are block elements like `div` they will just stack vertically without any CSS.

style.css
```diff
@@ -0,0 +1,281 @@
...
+.main-box-container {
+  background-color: #f4ece6;
+  box-shadow: rgba(0, 0, 0, 0.19) 0px 10px 20px, rgba(0, 0, 0, 0.23) 0px 6px 6px;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding-top: 1rem;
+  gap: 2.5rem;
+}
+img {
```
jonnywyatt, line 75: Using an element selector will style every image on the page, it's better to use a class to target just this image.

style.css
```diff
@@ -0,0 +1,281 @@
...
+  justify-content: center;
+  height: 3rem;
+}
+
+.hello-part {
+  display: flex;
+  flex-direction: column;
+  padding: 1rem;
+  gap: 2rem;
+  width: 25%;
```
jonnywyatt, line 112: This is being overridden by the styles in the media query using `max-width`

style.css
```diff
@@ -0,0 +1,281 @@
...
+}
+.hello-text {
+  font-size: 6.3rem;
+  font-family: "Poppins", sans-serif;
+}
+.role {
+  font-size: 1.6rem;
+}
+.p1 .p2 {
```
jonnywyatt, line 121: This is currently targetting `.p2` elements that are inside `.p1` elements, is that what you want?

index.html
```diff
@@ -1,21 +1,152 @@
...
+          />
+          <img
+            alt="Instagram"
+            src="https://static.wixstatic.com/media/01c3aff52f2a4dffa526d7a9843d46ea.png/v1/fill/w_19,h_19,al_c,q_85,usm_0.66_1.00_0.01,enc_auto/01c3aff52f2a4dffa526d7a9843d46ea.png"
+            style="width: 19px; height: 19px; object-fit: cover"
+          />
+        </div>
+      </div>
+      <!-- Please leave this in to credit the creators of this design -->
+      <p class="purpose2">
```
jonnywyatt, line 146: This text is repeated twice in the footer; if you can, try not to duplicate text and use CSS to control its styling

index.html
```diff
@@ -1,21 +1,152 @@
...
+          </p>
+        </div>
+      </div>
+    </main>
+    <footer>
+      <p class="purpose1">
+        This website design was created by Wix.com, and is used here for
+        strictly educational purposes.
+      </p>
+      <div class="footer-call">
```
jonnywyatt, line 110: The 3 contact elements (call, write, follow) are grouped together on larger screens. As we discussed on the call, it's ok to add a wrapper element around them to make it easier to achieve that styling.

jonnywyatt (commented):
There's some really good things here! Like the sticky header and the button styling.
Remember for mobile-first styling, use styles at the top of the stylesheet for small screens, then add a media query using `min-width` to change the styling for wider screens.
Also, if elements share styling, think about using a class for that and add it to all the similar elements. Eg a `.btn` class to apply the rounded corners to all buttons.

PR #220, opened 2022-11-18
LONDON-9-SAGHAR-HOSSEINMARDI-WEEK-4-RESPONSIVE-GRID-PROJECT

SagharHosseinmardi:


css/main.css
```diff
@@ -1,3 +1,213 @@
 /* Add your CSS here */
 
 /* Dont' forget to link this file to your HTML in the <head> */
+
+header {
+  display: grid;
```
jonnywyatt, line 6: This is quite a complicated way to style the header. If you think of it as two containers, one of them left-aligned with the name and 'project manager', and the other right-aligned with the menu, what's a simpler way to position them? If you can use simple ways to do layout it makes debugging problems much easier.

css/main.css
```diff
@@ -1,3 +1,213 @@
...
+  column-gap: 10px;
+  margin: 20px 15px 0 15px;
+  font-family: "Poppins", sans-serif;
+  color: black;
+  line-height: 2%;
+}
+
+main {
+  font-family: "Quicksand", sans-serif;
+  display: flex;
```
jonnywyatt, line 19: Remember, if you want elements to just stack vertically, block elements like `div` or `main` will do it naturally without any CSS, there's no need to use `flex`

css/main.css
```diff
@@ -1,3 +1,213 @@
...
+  background-color: #0150fd;
+  grid-column: 1;
+  grid-row: 1;
+}
+
+.toggle-menu {
+  float: right;
+}
+
+@media screen and (min-width: 900px) {
```
jonnywyatt, line 34: Great, using `min-width` like this means your styles work 'mobile-first' 👍

index.html
```diff
@@ -1,21 +1,142 @@
...
+  <body>
+    <header>
+      <div class="toggle-menu">
+        <hr class="line-one" />
+        <hr class="line-two" />
+        <hr class="line-three" />
+      </div>
+
+      <div class="blue-box"></div>
+      <h1>Maya Nelson</h1>
```
jonnywyatt, line 31: The `h1` should be the text that best describes this page, as distinct from the others in the site. So if all the pages are about Maya Nelson, then the text "Maya Nelson" won't be the best choice. Which heading on this page best describes what this page is about? That should be the h1.

css/main.css
```diff
@@ -1,3 +1,213 @@
...
+h2 {
+  grid-column: 1 / span 2;
+  grid-row: 2;
+  font-size: 1.2em;
+  font-style: normal;
+  font-weight: 100;
+}
+
+.card {
+  width: 20em;
```
jonnywyatt, line 74: Using `em` for width like this can be a bit confusing; do you know how this value is calculated? `px` or `rem` is simpler.

css/main.css
```diff
@@ -1,3 +1,213 @@
...
+  background-color: #f4ece6;
+}
+
+.profile-img {
+  border-radius: 50%;
+  width: 150px;
+  margin-top: 20px;
+}
+.name {
+  font-size: 1.7em;
```
jonnywyatt, line 88: Do you know how this font size is calculated? `px` or preferably `rem` are easier to work with and more predictable. `em` relies on the parent's font size so if this element moves, it will change.

css/main.css
```diff
@@ -1,3 +1,213 @@
...
+  font-family: Quicksand;
```
jonnywyatt, line 89: If you can add a fallback font, it will use that if the user's device doesn't have this custom font.
eg `font-family: Quicksand, sans-serif`

index.html
```diff
@@ -1,21 +1,142 @@
...
+        <img
+          src="AdobeStock_230407429.webp"
+          alt="woman-picture"
+          class="profile-img"
+        />
+        <p class="name">Maya<br />Nelson</p>
+        <hr class="card-line" />
+
+        <div class="card-buttons">
+          <button class="card-buttons1">RESUME</button>
```
jonnywyatt, line 52: These will link somewhere; so better to use `a`. It can still be styled like a button, but choose `a` if it's going to act like a link.

index.html
```diff
@@ -1,21 +1,142 @@
...
+            alt="instagram-icon"
+            class="instagram"
+          />
+        </div>
+      </section>
 
-<body>
+      <section class="info-paragraph">
+        <div class="paragraph-text">
+          <h3><strong>Hello</strong></h3>
```
jonnywyatt, line 82: Remember headings shouldn't jump levels, there is no h1 or h2 in the page so you can't jump straight to h3

index.html
```diff
@@ -1,21 +1,142 @@
...
+          <p class="paragraph">
+            I’m a great place for you to tell a story and let your users know a
+            little more about you.
+          </p>
+        </div>
+      </section>
+    </main>
+    <footer>
+      <div class="contact">
+        <p><strong>Call</strong></p>
```
jonnywyatt, line 99: the phone number should be the element after the `Call` heading, in the HTML, so a screenreader will read them one after the other. The problem is using grid to place the elements in a square. Instead think of 'call, write, follow' as 3 wrapper elements in a single row. Then inside each wrapper, there is a heading and an element below it with the phone number / email / icons.

css/main.css
```diff
@@ -1,3 +1,213 @@
...
+  margin: 20px 15px 0 15px;
+  font-family: "Poppins", sans-serif;
+  color: black;
+  line-height: 2%;
+}
+
+main {
+  font-family: "Quicksand", sans-serif;
+  display: flex;
+  flex-direction: column;
```
jonnywyatt, line 20: On wider screens, the 2 subelements (`.card` and `.info-paragraph`) sit side by side rather than stacked vertically. Have a think how you can do that in your media query using `flex-direction`

css/main.css
```diff
@@ -1,3 +1,213 @@
...
+  margin-top: auto;
+}
+
+.card-line {
+  width: 6.5em;
+  border-top: 2px solid blue;
+}
+
+.card-buttons1 {
+  width: 140px;
```
jonnywyatt, line 100: Both buttons share some styles eg width, margin, padding. How can you extract those shared properties to a class eg `.button` and use that class on both buttons? Then you have less duplication in your CSS and it's easier to maintain.

jonnywyatt (commented):
The layout on small screens looks good; can you add a media query to change the layout on wider screens, to match the design?

css/main.css
```diff
@@ -1,3 +1,213 @@
...
+  grid-column: 1 / span 2;
+  grid-row: 2;
+  font-size: 1.2em;
+  font-style: normal;
+  font-weight: 100;
+}
+
+.card {
+  width: 20em;
+  height: 28em;
```
jonnywyatt, line 75: Try to avoid setting heights on elements unless you have to, so that if the text size or the amount of text changes, the box will grow in height to accommodate it.

PR #224, opened 2022-11-18
London-9-Karleen-Richards-HTML/CSS-Responsive-Grid-Project-Week4

karleenmsrichards:


css/main.css
```diff
@@ -1,3 +1,861 @@
...
+    --white: #FFF;
+    --lylac: #f5e8ff;
+    --light-green: #E6FAC0;
+    --medium-green: #5E7D1F;
+    --dark-green: #394B2A;
+    --orange: #ec7a17;
+}
+
+body {
+    font-family: Raleway;
```
PeteLindsell: Its good practice to add a fallback font eg: `font-family: Raleway, sans-serif;`

css/main.css
```diff
@@ -1,3 +1,861 @@
...
+    font-size: .9275rem;
+    line-height: 1.8em;
+    word-spacing: .1em;
+    text-align: center;
+    margin: 0;
+    padding: 0;
+    display: grid;
+    grid-template-columns: repeat(auto-fit, minmax(1fr, 1fr));
+    grid-template-rows: repeat(1fr, 1fr);
```
PeteLindsell: These are both invalid and there for not having any effect

`grid-template-columns: repeat(auto-fit, minmax(1fr, 1fr));` setting the two parameters of minmax() does not do anything 

`grid-template-rows: repeat(1fr, 1fr);` the first parameter of repeat() should be a number, auto-fit, or auto-fill a fractional unit wont work

When you are trying things out it is good to remove anything that doesnt have the effect you are after

css/main.css
```diff
@@ -1,3 +1,861 @@
...
+}
+
+@media only screen and (min-width: 900px) {
+    .header {
+        z-index: 500;
+        display: flex;
+        flex-direction: row;
+        justify-content: space-evenly;
+        align-items: center;
+        position: sticky;
```
PeteLindsell, line 118: Well done on finding sticky it works well here to get the effect you are after

css/main.css
```diff
@@ -1,3 +1,861 @@
...
+
+    .section2-content-wrapper {
+        display: grid;
+        grid-gap: 50px;
+        grid-template-columns: repeat(3, 1fr);
+        grid-template-rows: repeat(1, 1fr); 
+        margin: 10px auto 0 30px;
+        padding: 0 30px;
+        max-width: 100%;
+    }
```
PeteLindsell: The combination of margin right auto and margin left 30px is breaking the centred layout `margin: 10px auto 0 30px;` on larger screens. Auto left and right can be used to centre items but you are already in a centred flex element so there is no need here

css/main.css
```diff
@@ -1,3 +1,861 @@
...
+        margin: 0;
+        padding: 0;
+    }
+
+    .section4-img3 {
+        width: 310px;
+        height: 390px;
+        margin: 0;
+        padding: 0;
+    }
```
PeteLindsell: This could be written as
`.section4-img1,
.section4-img2,
.section4-img3 {
        width: 310px;
        height: 390px;
        margin: 0;
        padding: 0;
    }`

css/main.css
```diff
@@ -1,3 +1,861 @@
...
+    object-fit: contain;
+    position: absolute;
+    bottom: 60px;
+    left: 137px;
+    -webkit-transform:rotate(18deg);
+    -moz-transform: rotate(18deg);
+    -ms-transform: rotate(18deg);
+    -o-transform: rotate(18deg);
+    transform: rotate(18deg);
+}
```
PeteLindsell: Good use of `transform: rotate();` here this section is stating to look good.
You dont need all the vendor prefixes here any more as all the major browsers now support transform without a prefix

PeteLindsell (commented):
You have chosen a difficult layout which is coming together really well with some good use of `z-index` `sticky` and `transform` to accomplish some trick effects. Well done!

I have made a few comments bellow.

karleenmsrichards:
Thank you. I will make the necessary adjustments.

PR #227, opened 2022-11-18
html/css set up

howard-ss:
html/css set up


index.html
```diff
@@ -4,15 +4,86 @@
...
+
+     <section class="h-feature">
+        
+          <img src="maya.jpeg" alt="maya">
+          <h2>Maya</h2>
+          <h2 >Nelson</h2>
+          <button class="one"><a id="left" href="#">RESUME</a></button> <button class="two" ><a id="right" href="#">PROJECTS</a></button><br>
+        
+        <div class="logo">
+          <img src="facebook-icon.svg" alt="facebook" width="20">
```
SallyMcGrath, line 33: excellent alt text here. Would these be links on a real page?

index.html
```diff
@@ -4,15 +4,86 @@
...
+        <input type="checkbox">Maya Nelson</h2>
+       <p>PROJECT MANAGER</p>
+     </section>
+
+     <section class="h-feature">
+        
+          <img src="maya.jpeg" alt="maya">
+          <h2>Maya</h2>
+          <h2 >Nelson</h2>
+          <button class="one"><a id="left" href="#">RESUME</a></button> <button class="two" ><a id="right" href="#">PROJECTS</a></button><br>
```
SallyMcGrath, line 30: Should these be buttons or should they be links _styled_ as buttons?

Buttons trigger an action (like in a form) and links navigate to a new location. What is happening in this interaction?

index.html
```diff
@@ -4,15 +4,86 @@
...
+       </div>
+     </section>
+
+        <section class="h-aside">
+          <div class="h-up">
+            <h3>Call</h3>   
+            <h3>Write</h3>
+          </div>
+          <div class="h-down">
+            <p>123-456-7890</p>
```
SallyMcGrath, line 65: Could this be a +tel link?

index.html
```diff
@@ -4,15 +4,86 @@
...
+            <p>info@mysite.com</p>
```
SallyMcGrath, line 66: Could this be a :mailto link?

css/main.css
```diff
@@ -1,3 +1,120 @@
...
+    font-family: 'Poppins';
+}
+
+input {
+    background: blue;
+}
+
+/*define grid*/
+.h-main {
+     display: grid;
```
SallyMcGrath, line 25: When you disable this ruleset in Devtools, what happens to the layout? Do you need these rules?

You can  disable style rules by unchecking the box next to them in [Devtools](https://developer.chrome.com/docs/devtools/css/) .

Screenshot 2022-11-18 at 17 34 08

index.html
```diff
@@ -4,15 +4,86 @@
...
+    <link rel="stylesheet" href="css/main.css">
     <title>Responsive grid project</title>
 </head>
 
 <body>
 
     <!-- Add your markup here -->
+<main class="h-main">
 
+     <section class="h-hero">
```
SallyMcGrath, line 19: Is this a hero? What does hero mean in the context of web design?

https://uxplanet.org/unsung-heros-of-ui-look-at-hero-images-in-web-design-881e04fe7bd2

SallyMcGrath (commented):
Thanks for sharing your work so far!

In general my advice is to forget about the whole page layout and take each component separately. Write out a list (or some cards on a project board) of all the components this page has and then build each one. You will probably find that the whole page layout is very simple after that.

This is called atomic design, and it's a common approach in front end.

Start by identifying each component on the page

- [ ] header
- [ ] header navigation
- [ ] header menu button
- [ ] profile card
- [ ] social links navigation

...etc!

Breaking the problem down into small, manageable steps, will make it easier to reason about.

index.html
howard-ss: These 4 logos need to link to real pages.

index.html
howard-ss: The two buttons link to new locations to separate pages.

index.html
howard-ss: There's no a +tel link as per original design.

index.html
howard-ss: No, this is not a :mailto link as per original design.

css/main.css
howard-ss: I didn't realise this function in Devtools, I need to be familiar with it and how to use it .
Thanks for letting me know it.

index.html
howard-ss: Thanks for emphasising 'hero' concept. 
As per original design, this section should be 'hero' .

howard-ss:
I need to re-design this website as per your 'to-the-point' comments. 
Many thanks!

index.html
SallyMcGrath: (This should be a mailto link)

index.html
SallyMcGrath: Right! Great answer. So these should be links, not buttons. 

Here you have buttons with links inside. This is invalid html. Buttons cannot contain links nor links contain buttons. This is because they are both interactive element types. This is [defined in the spec](https://www.w3.org/TR/2011/WD-html5-20110525/content-models.html#interactive-content) but I do not expect you to have read it!

tl:dr - don't put a link in a button

index.html
SallyMcGrath: I think it could be a tel link! 😉

css/main.css
SallyMcGrath: It's soooo useful. I use it all the time! Devtools are amazing. So helpful!

PR #236, opened 2022-11-19
London class 9_Turing - GayleNgozi_Thompson-Igwebike - HTML/CSS week 4

Gayle-Thompson-Igwebike:
working progress of resume website


SallyMcGrath:
👋 Hi @GayleNThompsonIgwebike 

Please open your PR using the template provided and the title requested. Thanks!

https://www.awesomecodereviews.com/pull-request-template/

SallyMcGrath:
(Thanks for updating!!)

css/main.css
```diff
@@ -1,3 +1,225 @@
 /* Add your CSS here */
+*{
 
+}
+
+body{
+    /* display: grid; */
```
SallyMcGrath: Whoops, some commented out code left in this PR. 

https://syllabus.codeyourfuture.io/guides/code-style-guide#dont-leave-lots-of-commented-out-code

SallyMcGrath:
```suggestion
```

css/main.css
```diff
@@ -1,3 +1,225 @@
...
+    /* grid-template-columns: ; */
```
SallyMcGrath:
```suggestion
```

css/main.css
```diff
@@ -1,3 +1,225 @@
...
+    /* min-width: auto; */
```
SallyMcGrath:
```suggestion
```

css/main.css
```diff
@@ -1,3 +1,225 @@
...
+  padding: 3px;
+}
+
+.socials-container img{
+  margin-top: 10px;
+  height: 25px;
+  width: 25px;
+}
+
+    /*
```
SallyMcGrath: And here, let's clean this up. If you would like to share this code to discuss it, make a separate commit with this alternative stuff. You can check out different versions of the same page.

index.html
```diff
@@ -4,14 +4,81 @@
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <meta http-equiv="X-UA-Compatible" content="ie=edge">
-    <!-- Add a link to your custom CSS here -->
+    <link rel="stylesheet" href="css/main.css">
```
SallyMcGrath: I think right here might be a good place to load some custom web fonts. The fonts specified in the description are both Google Fonts, so you can grab them here: https://fonts.google.com/

What does MDN say about custom web fonts? https://developer.mozilla.org/en-US/docs/Learn/CSS/Styling_text/Web_fonts

index.html
```diff
@@ -4,14 +4,81 @@
...
+      
+<main>
+        <div class="greyandtext">
+            <div class="grey-box"></div>
+        </div>
+
+       <div class="container">
+         <div class="card">
+            <br>
+            <img src="/designs/resumepic.jpeg" alt="myphoto">
```
SallyMcGrath: Is this alt text descriptive for someone who cannot see the picture? How would a screenreader read out the string "my-photo" and what is communicated by this?

index.html
```diff
@@ -4,14 +4,81 @@
...
+            <class id="firstname"><h2>GAYLE</h2></class> 
+            <class id="lastname"><h2>NGOZI</h2></class>
+
+            <div class="blue-line">
+            <hr size="3px" width="50px" color="blue"> </div>
+
+            <div class="title"><p>SOFTWARE DEVELOPER</p></div>
```
SallyMcGrath: Is there a semantic html element you might use to describe a title or heading?

index.html
```diff
@@ -4,14 +4,81 @@
...
+            <hr size="3px" width="50px" color="blue"> </div>
+
+            <div class="title"><p>SOFTWARE DEVELOPER</p></div> 
+
+            <div class="buttons">
+            <button class="resume">RESUME</button>
+            <button class="projects">PROJECTS</button>
+          </div>
+
+            <div class="socials-container">
```
SallyMcGrath: What might this group of pictures be doing? What is their function? Is there a more meaningful way to describe this function? Should these link to places? Is this a list?

index.html
```diff
@@ -4,14 +4,81 @@
...
+            <class id="firstname"><h2>GAYLE</h2></class> 
+            <class id="lastname"><h2>NGOZI</h2></class>
+
+            <div class="blue-line">
+            <hr size="3px" width="50px" color="blue"> </div>
+
+            <div class="title"><p>SOFTWARE DEVELOPER</p></div> 
+
+            <div class="buttons">
+            <button class="resume">RESUME</button>
```
SallyMcGrath: Should this be a button or a link styled as a button? What is the difference?

(Hint: links navigate you somewhere else in the DOM)

index.html
```diff
@@ -4,14 +4,81 @@
...
+                <img src="/designs/linked.png" alt="linkedin icon">
+                <img src="/designs/instagram-logo.png" alt="instagram icon"> 
+              </div>
+         </div>
+ 
+    <div class="main-text">
+           <p id=“greeting”><h1>Hello</h1></p>
+             <p id=“intro”>Here’s who I am & what I do</p>
+        <div class=“buttons2”>
+                    <button class="resume">RESUME</button>
```
SallyMcGrath: Should this be a button or a link styled as a button?

css/main.css
```diff
@@ -1,3 +1,225 @@
...
+    font-family: "poppins", sans serif;
+  }
+
+.topnav{
+    display: flex;
+    flex-direction: row;
+     
+}
+
+#myLinks a {
```
SallyMcGrath: In general, try to avoid the use of ids in your CSS. This isn't an absolute rule, but a helpful hint for managing specificity when your designs become more complex. Read around this idea: https://dev.to/clairecodes/reasons-not-to-use-ids-in-css-4ni4

css/main.css
```diff
@@ -1,3 +1,225 @@
...
+  margin:10px auto;
+  font-size: 15px;
+}
+
+
+.container{
+  width:100%;
+  display: grid;
+  grid-template-areas: " left right";
+  position:absolute;
```
SallyMcGrath: This looks pretty wild on my screen. In general, using position:absolute for layout like this creates more problems than it solves.

Try resizing your browser, and choosing different views in Devtools Device Mode. 

https://developer.chrome.com/docs/devtools/device-mode/

What happens to your layout? 😱 😂

css/main.css
```diff
@@ -1,3 +1,225 @@
...
+  height: 32em;
+  right: -20em;
+  gap: px;
+}
+
+.main-text{
+    width: 50%;
+    grid-area: right;
+    margin: auto;
+    position: absolute;
```
SallyMcGrath: I think this is all a lot more complicated than it needs to be, and I think the place it's going wrong is the grey box. Try removing the grey box and all this absolute positioning. 

Have a look into CSS linear-gradient. Here's a codepen you might find interesting. https://codepen.io/stefanjudis/pen/vYOjJdZ?editors=1100

css/main.css
```diff
@@ -1,3 +1,225 @@
...
+
+.blue-button {
+  width: 20px;
+  height: 20px;
+  background-color: #0050FF;
+  margin-top: 20px;
+  margin-left: 14px;
+}
+
+  .grey-box{
```
SallyMcGrath: I don't think you need this and I think it's sending you down a rabbit hole of positioning nightmares. Try deleting this entirely and using a linear-gradient background on the body.

css/main.css
```diff
@@ -1,3 +1,225 @@
...
+
+}
+
+
+.card{
+    grid-area: left;
+    display: flex;
+    flex-direction: column;
+    background-color: #F4ECE6;
+    right:-20em;
```
SallyMcGrath: I think you can drop these negative positions here too. Work on the inside of the card component only.

Imagine the page as like a bento box. https://www.google.com/search?q=bento+box

Your card is like one delicious salad for your bento box. Your card can't know what kind of screen it will show up on, just like your salad doesn't know what bento box it will be spooned into. 🍱 

The page layout rules belong on the page level. Don't try to position each component by hand;  instead define a layout and "pour" each component into the layout section.

css/main.css
```diff
@@ -1,3 +1,225 @@
...
+    margin-top: 60px;
+    /* height: 30em; */
+    /* margin-left: 10%; */
+    box-shadow: grey -3px 3px 8px;
+    }
+   
+img{
+    width: 120px;
+    height: 120px;
+    border-radius: 50%;
```
SallyMcGrath: nice circle!

Another great way to do this is with `clip-path`. Have a play https://bennettfeely.com/clippy/

css/main.css
SallyMcGrath: (This is a complicated but **very useful** concept. Don't feel dismayed if you don't get this straight away! Take this idea to study sessions and buddy calls and think about it together. )

css/main.css
```diff
@@ -1,3 +1,225 @@
...
+  width:7em;
+  margin:10px auto;
+  font-size: 15px;
+}
+
+
+.container{
+  width:100%;
+  display: grid;
+  grid-template-areas: " left right";
```
SallyMcGrath: Here's a fun idea. You could  remove the gap rule and instead define _empty_  columns in CSS grid like this:

`grid-template-areas: ". card . intro .";`

Then you could try setting rules about how those columns are sized something like this:

`grid-template-columns: 1fr minmax(auto, 375px) 20px 3fr 1fr`

Have a play around.

css/main.css
```diff
@@ -1,3 +1,225 @@
...
+  position:absolute;
+  top: 8rem;
+  padding: 120px 0 0 150px;
+  height: 32em;
+  right: -20em;
+  gap: px;
```
SallyMcGrath: This CSS unit is missing a value

SallyMcGrath (commented):
Hi Gayle.

Thanks for your work so far. I can see you've put a lot of effort in. I've made a TON of comments here. Don't try to "fix" them all, certainly not at once.  I would take a look through the HTML comments first, and then come back and take another shot at the layout when you have a bit of space.

You're doing great, Gayle. Onward!

Gayle-Thompson-Igwebike:
Wow....thank you very much for this epic review of my work and your directions Sally. I truly appreciate the effort you must have put into this. I did not see this before now and just today, I deleted all my work to start over again because I just wasn't getting why it wont behave- so this is very very useful. THANK YOU. I will start it again with all your comments now as a guide. You are AWESOME Sally!! @SallyMcGrath

